PR #34, opened 2025-10-31
fix: exit successfully on validation warnings

MeelahMe:


MeelahMe:
**What changed:**
- Validation now exits successfully when there are only warnings
- Tested locally: basic_transformation has 4 warnings, 0 errors → exit code 0

PR #39, opened 2026-01-12
synthefy plugin

ajanbekzat:
## Synthefy Forecasting Plugin for InfluxDB 3

This PR adds a new HTTP trigger plugin that integrates [Synthefy Forecasting API](https://synthefy.com) with InfluxDB 3 to enable on-demand time series forecasting.

### Features

- **On-demand forecasting** via HTTP requests
- **Multiple model support**: Works with Synthefy models (sfm-tabular, Migas-latest, Chronos2, Moirai2, TimesFM, etc.)
- **Multivariate forecasting**: Supports metadata fields as covariates
- **Tag filtering**: Filter time series by tags (location, device, etc.)
- **Automatic data writing**: Forecasts written back to InfluxDB using Line Protocol

### How It Works

1. Queries historical time series data from InfluxDB based on measurement, field, tags, and time range
2. Transforms data to Synthefy API v2 request format
3. Calls Synthefy Forecasting API to generate forecasts
4. Writes forecast results (with quantiles) back to InfluxDB

### Requirements

- Python 3.7+
- `pandas` and `httpx`/`requests` packages
- Synthefy API key ([create one here](https://console.synthefy.com/api-keys))

### Usage Example
```
curl -X POST http://localhost:8181/api/v3/engine/forecast \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "measurement": "temperature",
    "field": "value",
    "tags": "location=NYC",
    "time_range": "30d",
    "forecast_horizon": "7d",
    "model": "sfm-tabular",
    "api_key": "your-api-key",
    "metadata_fields": "humidity,pressure"
  }'
```


garylfowler:
Thank you for this submission; we will be reviewing soon.

synthefy/synthefy-forecasting/synthefy_forecasting.py
```diff
@@ -0,0 +1,893 @@
...
+            "required": false
+        },
+        {
+            "name": "model",
+            "example": "sfm-tabular",
+            "description": "Synthefy model to use (e.g., 'sfm-tabular', 'Migas-latest'). See README for supported models.",
+            "required": false
+        },
+        {
+            "name": "api_key",
```
Aliaksei-Kharlap: For new plugins, it is advisable to only pass sensitive data like api_key through request headers or environment variables

synthefy/synthefy-forecasting/README.md
```diff
@@ -0,0 +1,404 @@
...
+
+### HTTP trigger parameters
+
+| Parameter            | Type   | Default  | Description                                                       |
+|----------------------|--------|----------|-------------------------------------------------------------------|
+| `measurement`        | string | required | Source measurement containing historical data                     |
+| `field`              | string | "value"  | Field name to forecast                                            |
+| `tags`               | string | ""       | Tag filters, comma-separated (e.g., "location=NYC,device=sensor1") |
+| `time_range`         | string | "30d"    | Historical data window. Format: `<number><unit>` (e.g., "30d")    |
+| `forecast_horizon`   | string | "7d"     | Forecast duration. Format: `<number><unit>` or "<number> points"  |
```
Aliaksei-Kharlap: For better understanding, it is worth specifying the supported units (d, h) in the documentation

synthefy/synthefy-forecasting/README.md
```diff
@@ -0,0 +1,404 @@
...
+| `model`              | string | "sfm-tabular"| Synthefy model to use (e.g., "sfm-tabular", "Migas-latest") |
+| `api_key`            | string | required | Synthefy API key (create at [console.synthefy.com/api-keys](https://console.synthefy.com/api-keys) or set SYNTHEFY_API_KEY environment variable)   |
+| `output_measurement` | string | "{measurement}_forecast" | Destination measurement for forecast results          |
+| `metadata_fields`    | string | ""       | Comma-separated list of metadata field names to use as covariates |
+| `database`           | string | ""       | Database name for reading and writing data (optional)             |
```
Aliaksei-Kharlap: The documentation states that this parameter is also used to specify the database to read data from. However, the plugin's code doesn't support this logic, and triggers can only read data from the database in which they were created

synthefy/synthefy-forecasting/synthefy_forecasting.py
```diff
@@ -0,0 +1,893 @@
...
+    retry_count = 0
+    for attempt in range(max_retries):
+        try:
+            for builder in builders:
+                influxdb3_local.write_to_db(database, builder)
+            logger.info(
+                f"Forecasts written successfully to InfluxDB (attempt {attempt + 1})"
+            )
+            return
+        except Exception as e:
```
Aliaksei-Kharlap, line 618: Currently, the processing engine for the influxdb3_local.write_to_db and influxdb3_local.write functions does not throw errors when writing during trigger execution. The current architecture works as follows: during trigger execution, all records are accumulated in a buffer, and the write itself occurs after the trigger completes. If there are write errors (different field types, etc.), errors are raised at the InfluxDB level. This means that the trigger is unaware of whether the write completed successfully and will never receive an error for the write operation. To address this issue, support for the influxdb3_local.write_sync and write_sync_to_db functions was added in InfluxDB 3.8.2. The write occurs immediately, which allows for error detection. You can see usage examples in other plugins (amqp_subscriber, kafka_subscriber, mqtt_subscriber)

synthefy/synthefy-forecasting/synthefy_forecasting.py
```diff
@@ -0,0 +1,893 @@
...
+                body_str = request_body.decode("utf-8")
+                body_dict = json.loads(body_str)
+            except (UnicodeDecodeError, json.JSONDecodeError) as e:
+                logger.warning(f"Failed to decode/parse request body: {e}")
+                body_dict = {}
+        elif isinstance(request_body, str):
+            try:
+                body_dict = json.loads(request_body)
+            except json.JSONDecodeError as e:
+                logger.warning(f"Failed to parse JSON string: {e}")
```
Aliaksei-Kharlap: I think it's best to throw an error here and terminate the plugin. If there's an error in the request body, the plugin will continue executing implicitly with the settings in args, even though the user expects the behavior they specified in the request body

synthefy/synthefy-forecasting/synthefy_forecasting.py
```diff
@@ -0,0 +1,893 @@
...
+        elif isinstance(request_body, dict):
+            body_dict = request_body
+        elif request_body is None:
+            body_dict = {}
+        else:
+            logger.warning(f"Unexpected request_body type: {type(request_body)}")
+            body_dict = {}
+
+        # Merge query parameters, request body, and args (request body can override args)
+        merged_args = {**args, **body_dict, **query_parameters}
```
Aliaksei-Kharlap: Since the documentation doesn't say anything about using query parameters, it's worth removing them

synthefy/synthefy-forecasting/synthefy_forecasting.py
```diff
@@ -0,0 +1,893 @@
...
+except ImportError:
+    try:
+        import requests as httpx
+    except ImportError:
+        raise ImportError(
+            "Either 'httpx' or 'requests' must be installed for API calls"
+        )
+
+# Configure logging
+logging.basicConfig(
```
Aliaksei-Kharlap: For logging during plugin execution, it is recommended to use the built-in functions influxdb3_local.info influxdb3_local.warn and influxdb3_local.error. This allows you to see logs from the processing engine in the general logging stream and save them to an internal table

Aliaksei-Kharlap: It is also desirable to have a unique identifier for each plugin execution (other plugins use task_id: str = str(uuid.uuid4()) which is added to the message in all logs)

synthefy/synthefy-forecasting/README.md
```diff
@@ -0,0 +1,404 @@
...
+
+This plugin includes a JSON metadata schema in its docstring that defines supported trigger types and configuration parameters. This metadata enables the [InfluxDB 3 Explorer](https://docs.influxdata.com/influxdb3/explorer/) UI to display and configure the plugin.
+
+### HTTP trigger parameters
+
+| Parameter            | Type   | Default  | Description                                                       |
+|----------------------|--------|----------|-------------------------------------------------------------------|
+| `measurement`        | string | required | Source measurement containing historical data                     |
+| `field`              | string | "value"  | Field name to forecast                                            |
+| `tags`               | string | ""       | Tag filters, comma-separated (e.g., "location=NYC,device=sensor1") |
```
Aliaksei-Kharlap: When passing parameters via --trigger-arguments, avoid commas in the values ​​themselves, as parsing will fail. In the case of --trigger-arguments tags=room=Bedroom,room2=Bedroom,field=temp, parsing will return {"tags": "room=Bedroom", "room2": "Bedroom", "field": "temp"}, which is incorrect.
It is recommended to use a space (tags="room=Bedroom room2=Bedroom",field=temp)

synthefy/synthefy-forecasting/README.md
```diff
@@ -0,0 +1,404 @@
...
+| `time_range`         | string | "30d"    | Historical data window. Format: `<number><unit>` (e.g., "30d")    |
+| `forecast_horizon`   | string | "7d"     | Forecast duration. Format: `<number><unit>` or "<number> points"  |
+| `model`              | string | "sfm-tabular"| Synthefy model to use (e.g., "sfm-tabular", "Migas-latest") |
+| `api_key`            | string | required | Synthefy API key (create at [console.synthefy.com/api-keys](https://console.synthefy.com/api-keys) or set SYNTHEFY_API_KEY environment variable)   |
+| `output_measurement` | string | "{measurement}_forecast" | Destination measurement for forecast results          |
+| `metadata_fields`    | string | ""       | Comma-separated list of metadata field names to use as covariates |
```
Aliaksei-Kharlap: the same as for the tags field

synthefy/synthefy-forecasting/synthefy_forecasting.py
```diff
@@ -0,0 +1,893 @@
...
+    headers = {
+        "Content-Type": "application/json",
+        "X-API-Key": api_key,
+    }
+
+    logger.info(f"Calling Synthefy API: {endpoint}")
+    logger.debug(f"Request data: {json.dumps(request_data, indent=2)}")
+
+    try:
+        if hasattr(httpx, "post"):
```
Aliaksei-Kharlap: The httpx library always has a post method, the else branch is dead code

synthefy/synthefy-forecasting/README.md
```diff
@@ -0,0 +1,404 @@
...
+
+```bash
+pip install pandas httpx
+```
+
+Or using InfluxDB 3 package manager:
+
+```bash
+influxdb3 install package pandas
+influxdb3 install package httpx
```
Aliaksei-Kharlap: The plugin code tries to import requests if httpx is missing, but the documentation doesn't mention installing requests

synthefy/synthefy-forecasting/synthefy_forecasting.py
```diff
@@ -0,0 +1,893 @@
...
+    # Build WHERE clause
+    where_parts = [f"time >= now() - INTERVAL '{time_range}'"]
+
+    if tags:
+        tag_conditions = [f"{k} = '{v}'" for k, v in tags.items()]
+        where_parts.extend(tag_conditions)
+
+    where_clause = " AND ".join(where_parts)
+
+    query = f"""
```
Aliaksei-Kharlap, line 379: When composing SQL, it is recommended to wrap field and table names in single quotes, which will allow their names to be parsed correctly

synthefy/synthefy-forecasting/synthefy_forecasting.py
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/README.md
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/synthefy_forecasting.py
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/synthefy_forecasting.py
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/synthefy_forecasting.py
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/README.md
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/synthefy_forecasting.py
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/README.md
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/synthefy_forecasting.py
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/synthefy_forecasting.py
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/README.md
Aliaksei-Kharlap: Updated

synthefy/synthefy-forecasting/README.md
Aliaksei-Kharlap: Updated

Aliaksei-Kharlap (approved):

PR #40, opened 2026-02-02
chore(ci): Refactor README validation for flexible section structure

jstirnaman:
- Made parameter documentation detection more flexible to support categorized parameters
- Relaxed validation rules for troubleshooting, code overview, and optional sections
- Reformatted code to use consistent double-quote strings and improved spacing
- Updated examples to use the new `--path` parameter

### DIFF
--- a/influxdata/system_metrics/README.md
+++ b/influxdata/system_metrics/README.md
@@ -141,7 +141,7 @@ influxdb3 create trigger \
 
 ### Query collected metrics
 
-This plugin collects system metrics automatically. Query to view the collected data:
+This plugin collects system metrics automatically. After the trigger runs, query to view the collected data:
 
 ```bash
 influxdb3 query \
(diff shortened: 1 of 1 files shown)


influxdata/state_change/README.md
```diff
@@ -133,16 +137,54 @@ influxdb3 enable trigger --database mydb state_change_scheduler
...
+influxdb3 write \
+  --database sensors \
+  "temperature,location=office value=28.5"
+
+# Create and enable the trigger
+influxdb3 create trigger \
+  --database sensors \
+  --path "gh:influxdata/state_change/state_change_check_plugin.py" \
+  --trigger-spec "every:15m" \
+  --trigger-arguments "measurement=temperature,field_change_count=value:5,window=1h,senders=slack,slack_webhook_url=https://hooks.slack.com/services/..." \
```
caterryan: Are these examples supposed to be runnable by users? This one doesn't define all trigger arguments

jstirnaman:
> The biggest issue is that plugin logs are located in the trigger's database, not `_internal`. This is an issue across all documents.

I verified they're in the trigger's database **and** in `_internal`. We'll focus on the former here and document the latter in the system_data guide in docs-v2. I'll have a PR soon there.

caterryan:
> > The biggest issue is that plugin logs are located in the trigger's database, not `_internal`. This is an issue across all documents.
> 
> I verified they're in the trigger's database **and** in `_internal`. We'll focus on the former here and document the latter in the system_data guide in docs-v2. I'll have a PR soon there.

My mistake; you are correct. 
I was thinking of processing_engine_triggers and processing_engine_trigger_arguments which are not logged in the _internal tables and only in the trigger's database tables. This has been a point of confusion for users, so we want to avoid referencing the _internal tables in the plugin docs.

Issue #47, opened 2026-02-26
bug(import): Cannot resume import after server crash - stale state blocks recovery


caterryan:
## Summary
After server crash, orphaned imports show "already running" and cannot be resumed. Breaks crash recovery.

## Steps to Reproduce
1. Start an import with slow query interval (`query_interval_ms: 15000`)
2. While import is running, kill the server (`pkill -9 influxdb3`)
3. Restart the server
4. Check status - shows `overall_status: "running"` but nothing is running
5. Attempt to resume - receive error: "Import is already running"

## Expected Behavior

After server restart, resume should either:
- Detect the import is not actually running and allow resume
- Automatically detect stale `in_progress` state and restart from checkpoint (or beginning if no checkpoint)

## Actual Behavior

Resume is blocked with "already running" error. Import is stuck forever.

Aliaksei-Kharlap:
It has been fixed. Closing

PR #51, opened 2026-03-03
fix(import): crash recovery, stale detection, and InfluxDB v3 timestamps

Aliaksei-Kharlap:
Fix crash recovery, stale import detection, and InfluxDB v3 timestamp compatibility for the import plugin.

  Bug fixes

  - Crash recovery: On unhandled errors, the plugin now writes paused state before exiting, so the import can be resumed after fixing the issue. Previously, a crash left the import
   stuck in "running" forever.
  - Stale import detection: resume now checks if a "running" import is actually stale (last update >5 min ago) and allows resume instead of returning "already running". If no
  checkpoint exists, restarts from the beginning.
  - Missing import_state table on resume: Wrapped query in try/except — if the table doesn't exist yet (crashed before processing any tables), restarts the import.
  - InfluxDB v3 timestamps: paused_at_time extraction now handles both int nanoseconds (v1/v2) and ISO 8601 strings (v3).
  - Row-level errors: Invalid rows in convert_influxql_to_line_protocol are now skipped with a warning instead of crashing the entire batch.

  Improvements

  - Added completed state to import_pause_state — pause/cancel/resume now return errors for already-completed imports.
  - Extracted _write_import_pause_state() helper — replaced 8 duplicate LineBuilder patterns; fixed a missing time_ns() call in one resume_import path.
  - Removed duplicate check_pause_state() function (~48 lines) — replaced with existing get_import_pause_state().

  Docs

  - Updated README: crash recovery section, stale detection (5-min threshold), troubleshooting for "already running" after crash, completed state handling.


Aliaksei-Kharlap:
#47 I added a check for the last entry in the import_state table (written for each time interval for which data is collected (batch)). If the last entry was less than 5 minutes ago, is_running is returned; if it was more than 5 minutes ago, the migration can continue. @caterryan How long did it take you to submit a request to resume migration?

caterryan (approved):
lgtm

PR #61, opened 2026-03-19
refactor(import): credential handling only via headers

caterryan:
v1/2/3 credentials only provided via http headers
credentials are not persisted

- Remove credential fields from ImportConfig and plugin metadata to avoid storing sensitive data
- Add `extract_credentials()` function to read auth from request headers at runtime
- Update all credential-dependent functions to accept credentials dict parameter
- Fix header key casing to handle InfluxDB3's lowercase normalization
- Updated readme


garylfowler (approved):
Good idea and implementation.

Aliaksei-Kharlap (approved):

PR #62, opened 2026-03-20
feat(import): Remove org requirement from v2 API functions

caterryan:
## Summary
- Replace v2 `/api/v2/buckets` endpoint with InfluxQL `/query` using `SHOW DATABASES` for `get_source_databases_list`
- Replace v2 Flux API (`/api/v2/query`) with InfluxQL `/query` using `SHOW MEASUREMENTS` for `get_source_tables_list`, removing `source_org` requirement
- Clean up unused `source_org` variable

This makes both v2 functions consistent with v1 and with how `query_source_influxdb` already handles v2 queries, using the InfluxQL compatibility endpoint with DBRP mappings and token-based auth.

### DIFF
--- a/influxdata/import/import.py
+++ b/influxdata/import/import.py
@@ -3006,7 +3006,6 @@ def get_source_tables_list(
     source_url = body_data.get("source_url")
     influxdb_version = body_data.get("influxdb_version")
     source_database = body_data.get("source_database")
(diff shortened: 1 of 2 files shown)


garylfowler (approved):
Changes make sense and improves the user experience.

PR #71, opened 2026-04-01
refactor: flux query for v2 import

wdoconnell:
This PR completes the fixes and adjustments in https://github.com/influxdata/influxdb3_plugins/pull/70 to make the v2->v3 import plugin use Flux, rather than InfluxQL, for retrieving data. 

The motivation for the change is that querying data with InfluxQL in v2 [requires implementation of database and retention policy (DBRP) mappings to map v2 buckets to a v1 database and retention policy. ](https://docs.influxdata.com/influxdb/v2/query-data/influxql/). Such mappings are created [automatically](https://docs.influxdata.com/influxdb/v2/query-data/influxql/dbrp/#automatic-dbrp-mapping) in some but not all cases. As a result, using native flux for the v2 import avoids requiring the user to debug such mappings.

Where feasible, this uses the [Python Client Library](https://docs.influxdata.com/influxdb/v2/api-guide/client-libraries/python/) to reduce code duplication, and has been added as a requirement for use.

NOTE: As the UI using these plugins explains, it is not anticipated that this plugin should be used for large or complex migrations; dedicated migration tooling is currently in development.

### DIFF
--- a/influxdata/import/requirements.txt
+++ b/influxdata/import/requirements.txt
@@ -1,1 +1,2 @@
 requests
+influxdb-client
--- a/influxdata/import/import.py
+++ b/influxdata/import/import.py
@@ -94,15 +94,24 @@ from typing import (
 )
 
 import requests
+from influxdb_client import InfluxDBClient
+from influxdb_client.client.flux_table import FluxTable
 
 # Global HTTP session for connection pooling
 _http_session = None
+
+# Cache for v2 org IDs to avoid repeated API calls
+_v2_org_id_cache: Dict[Tuple[str, str], Optional[str]] = {}
+
+# Cache for v2 InfluxDB client instances
+_v2_client_cache: Dict[Tuple[str, str], InfluxDBClient] = {}
 
 # Configuration constants
 MAX_RETRIES = 5
 INITIAL_BACKOFF_SECONDS = 1
 MAX_BACKOFF_SECONDS = 16
 REQUEST_TIMEOUT_SECONDS = 30
+MS_PER_SECOND = 1000
 STALE_IMPORT_THRESHOLD_SECONDS = 300  # 5 minutes — if last import_state update is older, import is considered stale
 
 # Timestamp offset constants (for boundary adjustments)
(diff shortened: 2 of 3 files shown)


influxdata/import/requirements.txt
```diff
@@ -1 +1,2 @@
-requests
\ No newline at end of file
+requests
+influxdb-client
```
Aliaksei-Kharlap, line 2: we need to update README.md to include the new dependency influxdb3 install package influxdb-client

caterryan:
The Flux refactor introduced three issues causing v2 imports to import 0 rows:

  1. Missing field detection — schema.fieldKeys() can return empty results for v2, so analyze_column_schema found no field columns (every field_types.get(col) returned None). Fix: treat non-tag, non-time columns as fields with inferred
  types.
  2. Field renaming — check_influx_type_to_python_type returned False for the "unknown" type set by get_field_keys, causing fields to be renamed with type suffixes (e.g., temperature_float). Fix: return True for "unknown".
  3. Timestamp crash — parse_timestamp_to_nanoseconds didn't handle datetime objects (Flux client returns these instead of strings), hitting the else branch which tries int(timestamp * 1e9) on a datetime. Fix: add datetime handling.
  
  Fixes implemented in #72, but we can move them to this PR if you want

Aliaksei-Kharlap:
There's another error handling cases where the data contains the same name for a field and a tag. It throws this error:

 INFO influxdb3_py_api::system_py: processing engine: [2002c5a3-2f47-4013-b6b2-dc91eaa3edca] Querying data for 'airSensors2' from 2025-10-18 15:15:15.075166+00:00 to 2025-10-21 16:46:39.965268+00:00
 WARN influxdb3_py_api::system_py: processing engine: [2002c5a3-2f47-4013-b6b2-dc91eaa3edca] Flux query failed (attempt 1/5), retrying in 1s: (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json; charset=utf-8', 'Vary': 'Accept-Encoding', 'X-Influxdb-Build': 'OSS', 'X-Influxdb-Version': 'v2.8.0', 'X-Platform-Error-Code': 'invalid', 'Date': 'Wed, 01 Apr 2026 ', 'Transfer-Encoding': 'chunked'})
HTTP response body: b'{"code":"invalid","message":"runtime error @4:6-4:76: pivot: value \\"room\\" appears in a column key column, but a column named \\"room\\" already exists; consider renaming \\"room\\" to something else before pivoting"}'

wdoconnell:
Closing this for now. The flux cutover breaks at least the non-Cloud flow. Will put up a PR for a revert on the first stage of the plugin, which affects UI usage.

PR #72, opened 2026-04-01
fix: v2 flux import failing to import any rows

caterryan:
## Summary
- Fix v2 Flux-based import importing 0 rows for every table after the switch from InfluxQL to Flux queries
- Handle `schema.fieldKeys()` returning empty results by treating non-tag columns as fields with inferred types
- Add `datetime` object support in `parse_timestamp_to_nanoseconds` for Flux client timestamps
- Return `True` from `check_influx_type_to_python_type` for `"unknown"` type to preserve original field names

### DIFF
--- a/influxdata/import/import.py
+++ b/influxdata/import/import.py
@@ -1012,7 +1012,9 @@ def sample_data_density(
 
 
 def check_influx_type_to_python_type(influx_type: str, value) -> bool:
-    if influx_type == "string":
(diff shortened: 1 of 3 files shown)


caterryan:
Closing this for now. The flux cutover breaks at least the non-Cloud flow. Will put up a PR for a revert on the first stage of the plugin, which affects UI usage.

PR #74, opened 2026-04-01
feat(notifier): add process_writes WAL trigger entry point

caterryan:
## Summary
- Add `process_writes` WAL trigger entry point to the notifier plugin, enabling pub-sub notification dispatch when rows are written to a table
- Add new mode to `process_request` for env-var-based credential resolution via trigger args (`*_env` pattern)
- Batch async senders (Slack, Discord, HTTP) concurrently via `asyncio.gather()` in a single event loop
- Existing `process_request` legacy code path and all sender functions are completely untouched — no breaking changes

## Design

Add a WAL trigger entry point (`process_writes`) to the notifier plugin so it can send notifications both when it receives an HTTP request and when lines are written to a table. The table acts as a pub-sub notification channel: alert plugins publish rows, notifier triggers subscribe and dispatch. Secrets never touch the table — each notifier resolves its own credentials from environment variables, referenced by name through trigger args.

### Goals

- Support WAL triggers alongside the existing HTTP trigger
- Define a clear boundary (contract) between how data is received and how notifications are dispatched
- Keep secrets out of tables and trigger args — credentials live in env vars only
- Support N:M fan-out: multiple alert plugins write to a table, multiple notifier triggers fire on the same table
- No breaking changes to the existing `process_request` legacy code path

### Non-Goals

- Refactoring SMS/WhatsApp sender internals (DRY consolidation is out of scope)
- Adding new sender types
- Multi-file module split (plugin must remain a single deployable file)

### Architecture

The table is a **notification channel** (pub-sub message bus):

```
              PUBLISHERS                          SUBSCRIBERS
              (alert plugins write rows)          (notifier triggers fire on writes)

 ┌──────────────┐  ┌──────────────┐       ┌──────────────┐  ┌──────────────┐
 │  threshold   │  │   deadman    │       │    slack     │  │     sms      │
 │  alert       │  │   alert      │       │  notifier    │  │  notifier    │
 │  plugin      │  │   plugin     │       │  (trigger)   │  │  (trigger)   │
 └──────┬───────┘  └──────┬───────┘       └──────┬───────┘  └──────┬───────┘
        │                 │                      │                 │
        └────────┬────────┘                      └────────┬────────┘
                 │                                        │
                 ▼                                        ▼
        ┌────────────────────────────────────────────────────────┐
        │              notifications table                       │
        │              (the contract)                             │
        └────────────────────────────────────────────────────────┘
```

- **Publishers** (alert plugins) write rows containing the notification message and metadata
- **Subscribers** (notifier triggers) fire on every write and dispatch via their configured sender
- Fan-out is structural: attach N triggers to the same table → every row triggers N notifications
- Fan-in is natural: M alert plugins write to the same table → all messages reach all subscribers
- **Secrets never touch the table.** Each notifier owns its credentials via trigger args (env var names) + env vars (secret values)

### DIFF
--- a/influxdata/notifier/README.md
+++ b/influxdata/notifier/README.md
@@ -1,21 +1,27 @@
 # Notifier Plugin
 
-⚡ http  
+⚡ http, onwrite
 🏷️ notifications, webhooks, messaging, alerts 🔧 InfluxDB 3 Core, InfluxDB 3 Enterprise
 
 ## Description
 
-The Notifier Plugin provides multi-channel notification capabilities for InfluxDB 3, enabling real-time alert delivery through various communication channels. Send notifications via Slack, Discord, HTTP webhooks, SMS, or WhatsApp based on incoming HTTP requests. Acts as a centralized notification dispatcher that receives data from other plugins or external systems and routes notifications to the appropriate channels.
+The Notifier Plugin provides multi-channel notification capabilities for InfluxDB 3, enabling real-time alert delivery through various communication channels. Send notifications via Slack, Discord, HTTP webhooks, SMS, or WhatsApp — either via incoming HTTP requests or when rows are written to a table (WAL trigger). Acts as a centralized notification dispatcher that receives data from other plugins or external systems and routes notifications to the appropriate channels.
 
 ## Configuration
 
+### HTTP trigger (legacy mode)
+
 This HTTP plugin receives all configuration via the request body. No trigger arguments are required.
+
+### WAL trigger (onwrite mode)
+
+The WAL trigger receives `sender_type` and credential env var names via trigger arguments. No secrets are stored in the table or trigger args — all credentials live in environment variables, referenced by name.
 
 ### Plugin metadata
 
 This plugin includes a JSON metadata schema in its docstring that defines supported trigger types and configuration parameters. This metadata enables the [InfluxDB 3 Explorer](https://docs.influxdata.com/influxdb3/explorer/) UI to display and configure the plugin.
 
-### Request body parameters
+### Request body parameters (HTTP trigger)
 
 Send these parameters as JSON in the HTTP POST request body:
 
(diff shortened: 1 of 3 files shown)


garylfowler (approved):

influxdata/notifier/notifier_plugin.py
```diff
@@ -1,6 +1,6 @@
 """
 {
-    "plugin_type": ["http"]
+    "plugin_type": ["http", "onwrite"]
```
Aliaksei-Kharlap, line 3: we need to add a section with possible arguments for onwrite so that the UI can display

influxdata/notifier/README.md
```diff
@@ -109,15 +115,124 @@ influxdb3 create trigger \
...
+
+## WAL trigger arguments
+
+Each WAL trigger is configured with `sender_type` and the names of the environment variables that hold the credentials. The `*_env` keys tell the plugin which env vars to read — the trigger args contain env var **names**, not secret values.
+
+### Slack
+
+| Argument          | Required | Description                                              |
+|-------------------|----------|----------------------------------------------------------|
+| `sender_type`     | yes      | Must be `"slack"`                                        |
```
Aliaksei-Kharlap, line 164: I think sender_type should be in the contract table

Aliaksei-Kharlap: In this case, we'll be able to use a single trigger for several possible sending paths at once. Currently, the configuration is limited to one channel per trigger. To configure several channels at once, we need to pass not just one channel in the webhook_url_env arguments, but split them into separate ones (slack_webhook_url_env, sms_webhook_url_env, whatsapp_webhook_url_env, http_webhook_url_env). This will give us a single trigger that supports multiple channels simultaneously.

influxdata/notifier/notifier_plugin.py
```diff
@@ -255,6 +421,32 @@ def process_request(
     task_id: str = str(uuid.uuid4())
     influxdb3_local.info(f"[{task_id}] Starting request process")
 
+    # New mode: sender_type in args — credentials from env vars, body has notification_text only
+    if args and "sender_type" in args:
```
Aliaksei-Kharlap, line 425: It's the same here, I think it's better to leave sender_type in the request body

influxdata/notifier/notifier_plugin.py
Aliaksei-Kharlap: Since we now also require an args setting for request, we need to add a section for it as well. I honestly don't remember what it should be called for the HTTP trigger type; I'll check with the team

Aliaksei-Kharlap: must be http_args_config

influxdata/notifier/README.md
```diff
@@ -109,15 +115,124 @@ influxdb3 create trigger \
...
 
+## Table schema (WAL trigger contract)
+
+When using the WAL trigger, alert plugins write rows to a shared table. The notifier reads these columns from each row:
+
+| Column              | Type         | Required | Description                                                              |
+|---------------------|--------------|----------|--------------------------------------------------------------------------|
+| `time`              | timestamp    | yes      | When the alert fired (set automatically by InfluxDB on write)            |
+| `notification_text` | string field | yes      | The notification message to dispatch                                     |
+| `id`                | string field | no       | Unique ID for this alert instance (set by the alert plugin)              |
```
Aliaksei-Kharlap, line 149: I think we need to specify the ID somewhere else in the logs, not just in the case of a skip. So that we can track which specific message is being processed

influxdata/notifier/notifier_plugin.py
Aliaksei-Kharlap: I checked with the team, the UI doesn't support this parameter yet, but they'll fix it

PR #76, opened 2026-04-20
fix import plugin empty table status

Aliaksei-Kharlap:
If there's an empty table in the database to import, its status remains pending. This fix corrects this behavior by changing the status to completed.


caterryan (approved):
lgtm

PR #77, opened 2026-04-28
synchronize fields in statistics tables for broker plugins

Aliaksei-Kharlap:
I am synchronizing the fields in the statistics tables for broker plugins. This is necessary to standardize the statistics tables and simplify UI operations. These changes do not break existing statistics tables, the old fields will no longer be populated, and new ones will be added


caterryan (approved):
lgtm

PR #78, opened 2026-05-01
Add fake data generation plugin

colebow:
This is a fake data generation plugin that simulates some pretty basic bird data for visualization and testing purposes. It's more rich than the sample data we currently provide in the Explorer UI, as it creates a steady stream of data that goes on a random walk and continues to write as long as the trigger for the plugin is running. The README file contained within the PR contains more documentation on how to setup and use the plugin.

Note that this _is_ vibe-coded, but with some pretty heavy editing, oversight, and testing from me.


influxdata/fake_data_generator/fake_data_generator.py
```diff
@@ -0,0 +1,599 @@
...
+    return time.time_ns()
+
+
+def _positive_int(value: Any, default: int) -> int:
+    if value is None or value == "":
+        return default
+
+    parsed = int(value)
+    if parsed < 1:
+        raise ValueError("volume configuration values must be at least 1")
```
Aliaksei-Kharlap: I guess there should be something like 'bird_count and points_per_bird must be at least 1'

influxdata/library/plugin_library.json
```diff
@@ -219,6 +219,17 @@
                 "last_update": "2026-03-06",
                 "trigger_types_supported": ["scheduler"]
             },
+            {
+                "name": "Bird Tracking Simulator",
+                "path": "influxdata/fake_data_generator/fake_data_generator.py",
```
Aliaksei-Kharlap: Perhaps the plugin files should be named differently so that their names align with the logic behind creating the bird data?

colebow: I've renamed to just "Fake Data Generator," as I think the fact that it's fake bird telemetry is kind of secondary to the fact that it's generating data.

Aliaksei-Kharlap: There's one caveat: new fake data generation plugins will be added in the future, and distinguishing them will be a challenge. I could be wrong, but the UI likely searches for plugins by part of the path, and common file naming could cause problems. I'd need to check with @caterryan about this. I'll get back to you with an answer later

caterryan: we have another fake data generator plugin that is about to be released, and there may be more fake data generator plugins in the future.

to prevent users from being confused, this plugin needs to be renamed to closer match the type of fake data that it generates

caterryan: the name has been changed to bird data simulator

caterryan (approved):
lgtm

PR #81, opened 2026-05-01
fix: SQL injection in downsampler and basic_transformation plugins

### LINKED ISSUE #79: fix: SQL injection in downsampler and basic_transformation plugins
A security review of the InfluxData plugins flagged SQL injection vulnerabilities in two plugins that build queries from customer-controlled trigger args via Python f-strings instead of parameter binding.

  downsampler — Critical

  File: influxdata/downsampler/downsampler.py:911-931 (generate_tag_filter_clause)
  for key, values in tag_values.items():
      if len(values) == 1:
          sql_clause += f"AND\n\t\"{key}\" = '{values[0]}'\n"
      else:
          quoted_values = ", ".join(f"'{v}'" for v in values)
          sql_clause += f'AND\n\t"{key}" IN ({quoted_values})\n'

  tag_values comes from the customer's tag_values trigger arg with no quote escaping anywhere. A trigger configured with tag_values="room:Kitchen'; <injected SQL> --" produces:

  AND "room" = 'Kitchen'; <injected SQL> --'

  The HTTP-mode parser (parse_tag_values_for_http) also doesn't validate tag-key characters, so key is injectable too (the scheduler-mode parser does enforce a regex on keys — please make both paths
  consistent).

  basic_transformation — High

  File: influxdata/basic_transformation/basic_transformation.py:921-950 (generate_filter_clause)
  for field, op, val in filters:
      if isinstance(val, (int, float)):
          val_str = str(val)
      else:
          escaped = val.replace("'", "''")
          val_str = f"'{escaped}'"
      clauses.append(f'AND\n\t"{field}" {op} {val_str}\n')

  Values are escaped, but field (identifier) is not. A filter arg like 'foo" OR 1=1 --:=ignored' parses into field='foo" OR 1=1 --'which produces:

  AND "foo" OR 1=1 --" = 'ignored'

  Same plugin also has unvalidated measurement interpolation as FROM '{measurement}' in generate_query — a single quote in the measurement name escapes the literal.

  Recommended fix

  Both plugins already use parameter binding correctly elsewhere (e.g., influxdb3_local.query(query, {"measurement": measurement}) for information_schema.columns lookups). Extend that pattern to all
  customer-influenced values:

  - Literal values (tag values, filter values): bind via $param parameters, don't string-interpolate.
  - Identifiers (tag keys, field names, measurement names) — these can't be parameterized, so:
    a. Validate against ^[A-Za-z0-9_][A-Za-z0-9_-]*$, AND
    b. Cross-check against the live schema (SHOW TABLES / information_schema.columns) before interpolation.

  Threat-model context

In Hosted Enterprise, customer-supplied trigger args are untrusted input. The blast radius is bounded by the trigger's auth-token scope (the customer's own DB), so this isn't cross-tenant — but it does let a customer broaden a trigger's WHERE clause beyond what the trigger config implies, run arbitrary SELECTs against any table the trigger token can read, and potentially probe information_schema / system tables. Worth fixing on the same theory as any tenant-input → SQL path.

caterryan:
## Summary

Fixes SQL injection vulnerabilities flagged by security review in the downsampler and basic_transformation plugins where customer-controlled trigger args were interpolated into SQL queries via f-strings.

- **Literal values** (tag values, filter values): replaced f-string interpolation with `$param` parameter binding
- **Identifiers** (tag keys, field names, measurement names): validated with regex (`^[A-Za-z0-9_][A-Za-z0-9_-]*$` via `fullmatch`) and cross-checked against live schema; all identifier interpolation uses `quote_identifier()` which doubles embedded `"`
- **Filter operators**: validated against allowlist `{=, !=, >, <, >=, <=}` inside `generate_filter_clause`, closing a bypass where TOML config-file filters skipped the string-parser's operator validation
- **TOML config-file tag_values**: now schema-cross-checked in `parse_tag_values_for_scheduler`, matching the HTTP and string-parser paths

Closes #79

## Test plan

- [ ] Run downsampler with scheduler trigger + tag_values and verify queries execute correctly
- [ ] Run downsampler with HTTP trigger + tag_values and verify queries execute correctly
- [ ] Run downsampler with TOML config file + tag_values and verify tag keys are schema-validated
- [ ] Run basic_transformation with scheduled trigger + filters and verify queries execute correctly
- [ ] Verify that identifiers with special characters (e.g., `"`) are rejected by `validate_identifier`
- [ ] Verify that filter operators from TOML config files are validated (e.g., `OR 1=1 --` is rejected)
- [ ] Run `docker compose --profile test run --rm test-core-all` if available

### DIFF
--- a/influxdata/basic_transformation/basic_transformation.py
+++ b/influxdata/basic_transformation/basic_transformation.py
@@ -171,17 +171,6 @@ import math
 
 from pint import UnitRegistry
 
-IDENTIFIER_PATTERN = re.compile(r"^[A-Za-z0-9_][A-Za-z0-9_-]*$")
-
-
-def validate_identifier(name: str, kind: str = "identifier") -> None:
-    if not IDENTIFIER_PATTERN.fullmatch(name):
-        raise Exception(
-            f"Invalid {kind}: '{name}' "
-            f"(must start with a letter, digit, or underscore and contain only letters, digits, underscores, and hyphens)"
-        )
-
-
 def quote_identifier(name: str) -> str:
     return '"' + name.replace('"', '""') + '"'
 
@@ -957,7 +946,6 @@ def generate_filter_clause(
     params: dict = {}
 
     for idx, (field, op, val) in enumerate(filters):
-        validate_identifier(field, "filter field name")
         if op not in allowed_operators:
             raise Exception(
                 f"Invalid filter operator: '{op}' (must be one of {', '.join(sorted(allowed_operators))})"
@@ -995,7 +983,6 @@ def generate_query(
     Returns:
         tuple[str, dict]: A complete SQL query string and a dict of query parameters.
     """
-    validate_identifier(measurement, "measurement name")
     escaped_measurement = measurement.replace("'", "''")
 
     # SELECT clause
--- a/influxdata/basic_transformation/basic_transformation.py
+++ b/influxdata/basic_transformation/basic_transformation.py
@@ -171,6 +171,20 @@ import math
 
 from pint import UnitRegistry
 
+IDENTIFIER_PATTERN = re.compile(r"^[A-Za-z0-9_][A-Za-z0-9_-]*$")
+
+
+def validate_identifier(name: str, kind: str = "identifier") -> None:
+    if not IDENTIFIER_PATTERN.match(name):
+        raise Exception(
+            f"Invalid {kind}: '{name}' "
+            f"(must start with a letter, digit, or underscore and contain only letters, digits, underscores, and hyphens)"
+        )
+
+
+def quote_identifier(name: str) -> str:
+    return '"' + name.replace('"', '""') + '"'
+
 _OP_FUNCS = {
     ">": operator.gt,
     "<": operator.lt,
@@ -915,39 +929,48 @@ def generate_fields_string(tags: list[str], fields: list[str]) -> str:
         str: Formatted string with quoted field names separated by commas and newlines.
     """
     all_fields: list = tags + fields
-    return ",\n\t".join(f'"{field}"' for field in all_fields)
+    return ",\n\t".join(quote_identifier(field) for field in all_fields)
 
 
 def generate_filter_clause(
     filters: list[tuple[str, str, str | int | float]] | None,
-) -> str:
+    valid_columns: list[str] | None = None,
+) -> tuple[str, dict]:
     """
-    Generates the WHERE clause for filtering by field/operator/value tuples.
+    Generates the WHERE clause for filtering by field/operator/value tuples using parameter binding.
 
     Args:
         filters (list[tuple[str, str, str | int | float]] | None):
             A list of 3-tuples (field, operator, value) as returned by `parse_field_values`,
             or None.
+        valid_columns (list[str] | None): Known column names from the schema for cross-checking
+            filter field names. If provided, filter fields not in this list are rejected.
 
     Returns:
-        str: SQL WHERE clause string combining all filters with AND, or empty string if no filters.
+        tuple[str, dict]: SQL WHERE clause string and a dict of query parameters.
     """
     if not filters:
-        return ""
+        return "", {}
 
+    allowed_operators = {"=", "!=", ">", "<", ">=", "<="}
     clauses: list[str] = []
-    for field, op, val in filters:
-        # Determine whether to quote the value
-        if isinstance(val, (int, float)):
-            val_str = str(val)
-        else:
-            # escape single quotes in string values
-            escaped = val.replace("'", "''")
-            val_str = f"'{escaped}'"
+    params: dict = {}
 
-        clauses.append(f'AND\n\t"{field}" {op} {val_str}\n')
+    for idx, (field, op, val) in enumerate(filters):
+        validate_identifier(field, "filter field name")
+        if op not in allowed_operators:
+            raise Exception(
+                f"Invalid filter operator: '{op}' (must be one of {', '.join(sorted(allowed_operators))})"
+            )
+        if valid_columns is not None and field not in valid_columns:
+            raise Exception(
+                f"Filter field '{field}' does not exist in the measurement schema"
+            )
+        param_name = f"filter_val_{idx}"
+        params[param_name] = val
+        clauses.append(f'AND\n\t{quote_identifier(field)} {op} ${param_name}\n')
 
-    return "".join(clauses)
+    return "".join(clauses), params
 
 
 def generate_query(
@@ -957,9 +980,9 @@ def generate_query(
     tag_names: list[str],
     start_time: datetime,
     end_time: datetime,
-) -> str:
+) -> tuple[str, dict]:
     """
-    Builds an SQL query.
+    Builds an SQL query with parameter binding for filter values.
 
     Args:
         measurement: source measurement name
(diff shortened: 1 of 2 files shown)


influxdata/basic_transformation/basic_transformation.py
```diff
@@ -171,6 +171,20 @@
 
 from pint import UnitRegistry
 
+IDENTIFIER_PATTERN = re.compile(r"^[A-Za-z0-9_][A-Za-z0-9_-]*$")
+
+
+def validate_identifier(name: str, kind: str = "identifier") -> None:
```
Aliaksei-Kharlap: This regular expression prohibits the use of dots and other special characters in the names of fields, tags, and tables—even though Influx itself permits this (as can be seen in the documentation examples: https://docs.influxdata.com/influxdb3/enterprise/reference/naming-restrictions/#tag-keys-and-field-keys)

influxdata/basic_transformation/basic_transformation.py
```diff
@@ -915,39 +929,48 @@ def generate_fields_string(tags: list[str], fields: list[str]) -> str:
...
-            val_str = str(val)
-        else:
-            # escape single quotes in string values
-            escaped = val.replace("'", "''")
-            val_str = f"'{escaped}'"
+    params: dict = {}
 
-        clauses.append(f'AND\n\t"{field}" {op} {val_str}\n')
+    for idx, (field, op, val) in enumerate(filters):
+        validate_identifier(field, "filter field name")
```
Aliaksei-Kharlap: This check is redundant, as we perform a subsequent check: `if valid_columns is not None and field not in valid_columns`. This verifies that the field name matches the data defined in the schema

influxdata/basic_transformation/basic_transformation.py
```diff
@@ -970,25 +993,29 @@ def generate_query(
         end_time:   UTC datetime for WHERE time < ...
 
     Returns:
-        A complete SQL query string.
+        tuple[str, dict]: A complete SQL query string and a dict of query parameters.
     """
+    validate_identifier(measurement, "measurement name")
```
Aliaksei-Kharlap: The check here is also redundant, as the code follows the following logic: it retrieves the `measurement` from `args`, then uses it in `get_fields_names`—specifically within `influxdb3_local.query(query, {"measurement": measurement})`—and if an error occurs, the plugin's execution terminates at that point

Aliaksei-Kharlap: I think we can generally remove the `validate_identifier` function here

influxdata/downsampler/downsampler.py
```diff
@@ -326,6 +340,14 @@ def parse_tag_values_for_scheduler(
     # Use config file
     if args["use_config_file"]:
         if isinstance(tag_values, dict):
+            tag_names: list = get_tag_names(influxdb3_local, source_measurement, task_id)
+            for tag_name in list(tag_values.keys()):
+                validate_identifier(tag_name, "tag name")
```
Aliaksei-Kharlap: Further down, there is a check to ensure inclusion in the list of tags obtained from the schema, so this check is redundant

influxdata/downsampler/downsampler.py
```diff
@@ -390,6 +412,7 @@ def parse_tag_values_for_http(
     if tag_value_filters is not None:
         tag_names: list = get_tag_names(influxdb3_local, source_measurement, task_id)
         for tag_name in list(tag_value_filters.keys()):
+            validate_identifier(tag_name, "tag name")
```
Aliaksei-Kharlap: the same situation

influxdata/downsampler/downsampler.py
```diff
@@ -904,31 +929,46 @@ def generate_group_by_string(tags_list: list):
...
     if tag_values is None:
-        return ""
+        return "", {}
 
     sql_clause: str = ""
+    params: dict = {}
+    param_idx: int = 0
+
     for key, values in tag_values.items():
+        validate_identifier(key, "tag name")
```
Aliaksei-Kharlap: By this time, the parsers had already validated the values

influxdata/downsampler/downsampler.py
```diff
@@ -953,14 +993,18 @@ def build_downsample_query(
         end_time:   UTC datetime for WHERE time < ...
 
     Returns:
-        A complete SQL query string.
+        tuple[str, dict]: A complete SQL query string and a dict of query parameters.
     """
+    validate_identifier(measurement, "measurement name")
```
Aliaksei-Kharlap: The table names have already been validated in the `parse_source_and_target_measurement` function

influxdata/basic_transformation/basic_transformation.py
caterryan: Valid — InfluxDB allows dots, spaces, and other special characters in identifiers that this regex rejects. Removed `validate_identifier` and `IDENTIFIER_PATTERN` from both plugins in 6b2bd92. Injection prevention is now handled by `quote_identifier()` (doubles embedded double-quotes) and `$param` binding for literal values.

influxdata/basic_transformation/basic_transformation.py
caterryan: Agreed — removed in 6b2bd92. The schema cross-check via `valid_columns` plus `quote_identifier()` escaping covers this.

influxdata/basic_transformation/basic_transformation.py
caterryan: Agreed — removed in 6b2bd92. The upstream `get_fields_names` call validates the measurement exists, and `escaped_measurement.replace("\u0027", "\u0027\u0027")` handles the FROM clause.

caterryan: Done — removed `validate_identifier` and `IDENTIFIER_PATTERN` from both plugins in 6b2bd92.

influxdata/downsampler/downsampler.py
caterryan: Agreed — removed in 6b2bd92. The schema cross-check against `get_tag_names()` remains (added in this PR for the TOML path).

influxdata/downsampler/downsampler.py
caterryan: Removed in 6b2bd92. `quote_identifier()` still escapes the key at this layer as defense-in-depth.

influxdata/basic_transformation/basic_transformation.py
caterryan: Removed in 6b2bd92. Schema cross-check via valid_columns plus quote_identifier() escaping covers this.

influxdata/downsampler/downsampler.py
caterryan: Removed in 6b2bd92.

influxdata/downsampler/downsampler.py
caterryan: Removed in 6b2bd92. parse_source_and_target_measurement handles schema validation; the FROM clause uses single-quote escaping.

caterryan: Removed in 6b2bd92. parse_source_and_target_measurement handles schema validation; the FROM clause uses single-quote escaping.

caterryan: Removed in 6b2bd92.

Aliaksei-Kharlap (approved):

PR #89, opened 2026-05-17
[2] add toml file format check, unify file reading

Aliaksei-Kharlap:
add toml file format check, unify file reading for all plugins


caterryan (approved):
lgtm

PR #91, opened 2026-05-17
[4] remove default credentials, add auth_mechanism arg

Aliaksei-Kharlap:
remove default credentials, add auth_mechanism arg


caterryan (approved):
lgtm

PR #93, opened 2026-05-25
Akharlap/improve plugins

Aliaksei-Kharlap:
This PR fixes all the comments for plugins from the security team with a level of medium and above.
Affected plugins:
- Downsampler
- Basic Transformation
- US Weather Sampler
- MQTT Subscriber
- Kafka Subscriber
- AMQP Subscriber
- OPC UA

Also added toml/json file type validation for other plugins


caterryan (approved):
thanks

PR #110, opened 2026-07-08
add influxdata-plugin-utils package

Aliaksei-Kharlap:
# Add draft for shared helpers package

Introduces `test_package` (working name) — a reusable package of shared helpers for InfluxDB 3 plugins. The goal is to remove duplicated logic across plugins and provide a common set of utilities.

The package is built with **hatchling** and currently has a single dependency: **dynaconf**.

## Modules

| Module | Purpose |
| --- | --- |
| `config` | Dynaconf-backed `load_plugin_config()` with validation, `resolve_plugin_dir()` / `resolve_path()`, and re-exported `Validator` |
| `introspection` | Schema introspection helpers (`get_table_names()`, `get_tag_names()`, `get_field_names()`) and windowed queries (`query_window()`) |
| `parsing` | Value parsers: `parse_timedelta()`, `parse_int()`, `parse_bool()`, `parse_delimited_list()`, `parse_key_value()` |
| `cache` | `cached()` — TTL cache wrapper over `influxdb3_local.cache` |
| `write` | Line protocol builders (`build_line()`, `build_line_typed()`, `add_field_with_type()`) and resilient `write_data()` with batching and retries |

## Key features

  - Unified config loading with clear source precedence (low → high): explicitly named env vars (via env_keys) → engine args → TOML file. Layers merge per key rather than replacing each other wholesale, and TOML is parsed into native types with no manual string handling. The active layers are selectable via source (merge / args / toml).
  - Parameter validation via Validator (defaults, bounds, required fields, casting).
  - Resilient writes with batching and retries enabled by default, and opt-out support.
- TOML values are parsed into native types, removing the need for manual string conversion.

- Parameter validation through `Validator`:
  - defaults
  - bounds
  - required fields
  - type casting

- Resilient writes:
  - batching enabled by default
  - retries enabled by default
  - opt-out support when needed

## Example

`examples/simple_downsampler/` contains a scheduled plugin demonstrating usage of the full package:

- config loading and validation
- cached schema introspection
- windowed queries
- batched and retried writes


influxdata-plugin-utils/examples/simple_downsampler/simple_downsampler.py
caterryan, line 1: I don't think we need the examples

caterryan (approved):

influxdata-plugin-utils/examples/simple_downsampler/simple_downsampler.py
Aliaksei-Kharlap: Okay, I'll delete it

PR #113, opened 2026-07-09
feat(signal-generator): add simulated gaps

caterryan:
## Summary
- add configurable simulated outage gaps to signal generator output
- document default gap behavior, disable flag, and upgrade impact
- bump signal generator to 0.3.0 and update registry metadata

## Validation
- python3 -m pytest influxdata/signal_generator/test_signal_generator.py
- manually validated on influxdb:3-enterprise 3.10.2 with --without-auth and --plugin-dir=/plugins
- observed gap logs showing generated points removed and timestamp holes matching configured gap windows
- verified gap_enabled=false control trigger kept continuous 10 Hz output


influxdata/signal_generator/GAP_DESIGN.md
Aliaksei-Kharlap, line 1: I don't think there's any need to add files with plans to the repository, same point for GAP_IMPLEMENTATION_PLAN.md

Aliaksei-Kharlap (approved):

influxdata/signal_generator/GAP_DESIGN.md
caterryan: Whoops. Good catch

PR #115, opened 2026-07-13
Add Nori tabular regression plugin

ajanbekzat:
## Summary

`nori_regression` is a new Processing Engine plugin that predicts a numeric field in an InfluxDB 3 measurement from **other columns on the same rows** with **Nori**, Synthefy's in-context-learning tabular regression model, served through the Baseten gateway. It trains on the rows where the target field is present and predicts (imputes) the rows where it is null, writing the predicted values back as a new measurement. It runs as a **scheduled** trigger (on an interval) or an **HTTP** trigger (on-demand).

## What Nori is (for reviewers not familiar with it)

Nori is a tabular regression foundation model. You send it labeled feature rows (`X_train`, `y_train`) and query rows (`X_test`) in a single request, and it predicts a value for each query row in one forward pass, with no training or fine-tuning. It has no notion of time or row ordering: it only sees numeric feature rows.

## How the plugin works

1. Reads a window of the measurement: the target `field` plus the `feature_fields` columns, filtered by `tags`.
2. Splits rows into the training set (target present) and the prediction set (target null, all features present).
3. Sends `X_train`/`y_train`/`X_test` to Nori (`task: "regression"`).
4. Writes each prediction back at its own row's timestamp to `<measurement>_regressed`.

Typical uses: backfill a field that dropped out, impute a metric from correlated ones (for example, predict `pressure` from `temp` and `humidity`), or derive an expensive-to-measure field from cheaper ones recorded alongside it.

## Trigger types

- **Scheduled** (`process_scheduled_call`): run on an interval; re-imputes any rows still missing the target each tick.
- **HTTP** (`process_request`): on-demand at `/api/v3/engine/<path>`; supports an explicit `start_time`/`end_time` window.

## Configuration highlights

- `measurement`, `field`: the source table and the target field to predict (required).
- `feature_fields`: the feature columns (X), `.`/`,`-separated (required).
- `window`, or `start_time` + `end_time`: which rows to read.
- `tags`: filter to a single series (required if the measurement holds several series).
- `model`: the Nori gateway slug (`synthefy/nori` default, or `synthefy/nori-30m`).
- `output_measurement`, `target_database`, `dry_run`, `min_history`, `config_file_path`.

Full parameter tables are in the README.

## Authentication

The Nori gateway key is a secret and is **never** read from trigger arguments or the request body (both are logged). It is resolved from, in order: an incoming request header (`X-Nori-Api-Key`, or `Authorization: Api-Key` on the HTTP trigger), then the `NORI_API_KEY` environment variable on the InfluxDB host. Keys are obtained from the Synthefy console (https://console.synthefy.com/); a single key works for all Nori model variants.

One InfluxDB-specific note for the HTTP trigger: InfluxDB's own HTTP layer parses the `Authorization` header and only accepts the Bearer/Token/Basic schemes, so `Authorization: Api-Key` is rejected before the plugin runs. Use the custom `X-Nori-Api-Key` header instead. This is documented in the README.

## Testing

Verified end-to-end against `influxdb:3-core` v3.10.1 with a production key:

- On a `sensors` table where `pressure = 1000 + 0.5*temp - 0.3*humidity + noise(sd 0.8)`, the plugin imputed the 24 rows missing `pressure` from `temp` and `humidity`. Predictions matched the true relationship to within the training noise (absolute errors 0.02 to 0.71).
- A multi-series regression without a qualifying `tags` filter fails loud (a would-collide guard) rather than silently dropping values; a wrong or missing key returns a clean `failed` rather than crashing.

## Conformance to repo conventions

- `scripts/validate_readme.py --plugins nori_regression`: 0 errors.
- `scripts/validate_influxdb3_syntax.py --readme influxdata/nori_regression/README.md`: 0 errors, all CLI and API usage validated.
- Follows the `organization/plugin_name/` layout with the JSON metadata docstring, README, `manifest.toml`, and `requirements.txt`, plus a `plugin_library.json` entry (author Synthefy).
- Writes go through `write_sync`/`write_sync_to_db` (InfluxDB 3.8.2+, the same pattern as the `mqtt_subscriber` and `synthefy_forecasting` plugins) so write errors surface during trigger execution instead of being swallowed by the buffered writer.
- The Nori API key is read only from a request header or the `NORI_API_KEY` environment variable, never from trigger arguments or the request body.

## Files added

- `influxdata/nori_regression/nori_regression.py`: the plugin (metadata docstring plus implementation).
- `influxdata/nori_regression/README.md`: usage, configuration, and troubleshooting.
- `influxdata/nori_regression/manifest.toml`: packaging metadata.
- `influxdata/nori_regression/requirements.txt`: Python dependencies.
- `influxdata/library/plugin_library.json`: registry entry (the plugin lives in the `influxdata/` namespace, authored by Synthefy).

## Limitations

- One series per run; run multiple triggers for multiple series (the plugin fails loud if a run resolves to more than one series). Multi-series imputation in a single run is a possible enhancement.
- Imputes only rows where the target is null; it does not overwrite existing values.
- Prediction quality depends on how well the feature columns explain the target.
- Each call is a billed gateway request; a larger window sends more rows and tokens.

### DIFF
--- a/influxdata/library/plugin_library.json
+++ b/influxdata/library/plugin_library.json
@@ -341,6 +341,17 @@
                 "trigger_types_supported": ["http"]
             },
             {
+                "name": "Nori Regression",
+                "path": "influxdata/nori_regression/nori_regression.py",
+                "description": "[BETA] Requires InfluxDB 3.8.2+. Predict a numeric field in an InfluxDB 3 measurement from other columns with Nori, Synthefy's in-context-learning tabular regression model, via the Baseten gateway. Trains on the rows where the target field is present and predicts (imputes) the rows where it is null, writing the predictions back to InfluxDB. Scheduled and HTTP triggers; the model gateway slug is configurable (synthefy/nori or synthefy/nori-30m).",
+                "author": "Synthefy",
+                "docs_file_link": "https://github.com/influxdata/influxdb3_plugins/blob/main/influxdata/nori_regression/README.md",
+                "required_plugins": [],
+                "required_libraries": ["requests", "pandas", "numpy"],
+                "last_update": "2026-07-15",
+                "trigger_types_supported": ["scheduler", "http"]
+            },
+            {
                 "name": "SageMaker Inference",
                 "path": "influxdata/sagemaker/sagemaker.py",
                 "description": "[BETA] Requires InfluxDB 3.8.2+. Runs scheduled inference against Amazon SageMaker endpoints using recent InfluxDB 3 rows, supports multiple request and response shapes, and writes prediction results back to InfluxDB.",
--- a/influxdata/library/plugin_library.json
+++ b/influxdata/library/plugin_library.json
@@ -343,12 +343,12 @@
             {
                 "name": "Nori Regression",
                 "path": "influxdata/nori_regression/nori_regression.py",
-                "description": "[BETA] Requires InfluxDB 3.8.2+. Predict a numeric field in an InfluxDB 3 measurement from other columns with Nori, Synthefy's in-context-learning tabular regression model, via the Baseten gateway. Trains on the rows where the target field is present and predicts (imputes) the rows where it is null, writing the predictions back to InfluxDB. Scheduled and HTTP triggers; the model gateway slug is configurable (synthefy/nori or synthefy/nori-30m).",
+                "description": "[BETA] Requires InfluxDB 3.8.2+. Predicts a numeric field from other columns on the same rows with Nori, Synthefy's in-context tabular regression model, via the Synthefy inference gateway. Trains on the rows where the target is present and imputes the rows where it is null, writing predictions back to InfluxDB. Scheduled and HTTP triggers; the model slug is configurable (synthefy/nori-30m or synthefy/nori-6m).",
                 "author": "Synthefy",
                 "docs_file_link": "https://github.com/influxdata/influxdb3_plugins/blob/main/influxdata/nori_regression/README.md",
                 "required_plugins": [],
-                "required_libraries": ["requests", "pandas", "numpy"],
-                "last_update": "2026-07-15",
+                "required_libraries": ["influxdata-plugin-utils>=0.2.0", "requests"],
+                "last_update": "2026-07-30",
                 "trigger_types_supported": ["scheduler", "http"]
             },
             {
--- a/influxdata/nori_regression/README.md
+++ b/influxdata/nori_regression/README.md
@@ -1,0 +1,321 @@
+# Nori Regression Plugin
+
+⚡ scheduled, http 🏷️ regression, tabular, machine-learning 🔧 InfluxDB 3 Core, InfluxDB 3 Enterprise
+
+Predict a numeric field in an InfluxDB 3 measurement from other columns with **Nori**, Synthefy's
+in-context-learning tabular regression model, called through the Synthefy inference gateway. The
+plugin reads a window of rows, trains on the rows where the target field is present, predicts the
+rows where it is null (imputation / backfill), and writes the predicted values back into InfluxDB.
+
+## Description
+
+Nori is a tabular regression foundation model: you give it labeled feature rows (`X_train`,
+`y_train`) and query rows (`X_test`) in a single request, and it predicts a value for each query row
+in one forward pass, with no training or fine-tuning.
+
+This plugin applies Nori to an InfluxDB measurement. You choose a target field and a set of feature
+columns; the plugin uses the rows where the target is present as the in-context training set and
+predicts the target for the rows where it is null, writing each prediction back at its own row's
+timestamp. It is plain tabular regression: Nori sees only the feature columns you name, with no time
+or ordering assumptions.
+
+Typical uses:
+
+- Backfill a field that dropped out (a sensor went offline while its neighbors kept reporting).
+- Impute a missing metric from correlated ones (for example, predict `pressure` from `temperature`
+  and `humidity`).
+- Derive a field that is expensive to measure directly from cheaper ones recorded alongside it.
+
+Key features:
+
+- **In-context tabular regression**: no training step; the recent labeled rows are the context.
+- **Imputation / backfill**: predicts the rows where the target is null and writes them back.
+- **Scheduled or on-demand**: run on an interval (scheduled trigger) or via an HTTP endpoint.
+- **Tag filtering**: operate on a single series selected by tags.
+- **Writes results back** to InfluxDB as a new measurement using line protocol.
+
+## Configuration
+
+### Plugin metadata
+
+This plugin includes a JSON metadata schema in its docstring that declares the supported trigger
+types (`scheduled`, `http`) and their arguments, so the InfluxDB 3 Explorer UI can render a
+configuration form.
+
+### Authentication for the Nori gateway
+
+The Nori gateway API key is a secret and is **never** read from trigger arguments or the request
+body (both are logged). It is resolved, in order:
+
+1. an incoming `X-Nori-Api-Key: <key>` request header (HTTP trigger only), then
+2. the `NORI_API_KEY` environment variable set on the InfluxDB host (required for the scheduled
+   trigger).
+
+The key is intentionally **not** accepted in the `Authorization` header: InfluxDB parses
+`Authorization` for its own request authorization, so a key placed there never reaches the plugin.
+Use the custom `X-Nori-Api-Key` header instead.
+
+Get a Nori API key from the [Synthefy console](https://console.synthefy.com/). A single key works
+for all Nori model variants (see [Supported models](#supported-models)), so you do not need a
+separate key per variant. This plugin does not create keys.
+
+### Scheduled trigger parameters
+
+| Parameter | Default | Description |
+|---|---|---|
+| `measurement` | (required) | Source measurement (table) to read from. |
+| `field` | (required) | The numeric field to predict (the regression target). The plugin trains on rows where it is present and predicts the rows where it is null. |
+| `feature_fields` | (required) | Numeric feature columns (X) used to predict `field`, **space-separated** (e.g. `temp humidity`). Use spaces, not commas (`--trigger-arguments` splits argument pairs on commas) and not dots (field names may contain a `.`). In the HTTP JSON body it may also be a list. |
+| `window` | `30d` | Time window of rows to read. Units: `s,min,h,d,w`. |
+| `tags` | (none) | Filter to a single series. Format: `key:val key2:val2` (space-separated pairs). Required if the measurement holds more than one series (see [Notes](#notes)). |
+| `model` | `synthefy/nori` | The Nori gateway slug to call. Your API key must be granted this slug. |
+| `gateway_url` | `https://inference.baseten.co/predict` | Nori gateway endpoint. |
+| `output_measurement` | `<measurement>_regressed` | Measurement to write predictions to. |
+| `target_database` | (trigger db) | Write predictions to a different database. |
+| `dry_run` | `false` | If `true`, log predictions but do not write them. |
+| `min_history` | `50` | Minimum labeled rows (target present) required to train; skip below this. |
+
+### HTTP trigger parameters
+
+All scheduled parameters are accepted (in the JSON request body or as trigger arguments), plus:
+
+| Parameter | Default | Description |
+|---|---|---|
+| `start_time` | (none) | ISO start of the window. Overrides `window`. |
+| `end_time` | (none) | ISO end of the window. |
+
+Scalar values in the JSON body override trigger arguments. A `tags` object may be sent as JSON
+(`{"tags": {"site": "A"}}`), and `feature_fields` may be sent as a JSON list
+(`{"feature_fields": ["temp", "humidity"]}`) or a space-separated string.
+
+## Requirements
+
+### Dependencies
+
+- `requests`
+- `pandas`
+- `numpy`
+
+### Installation steps
+
+1. Install the Python dependencies into the InfluxDB 3 Processing Engine environment:
+
+   ```bash
+   influxdb3 install package requests pandas numpy
+   ```
+
+2. Reference the plugin directly from this repository with the `gh:` prefix (the form used in the
+   examples below): `--path "gh:influxdata/nori_regression/nori_regression.py"`. Alternatively, copy
+   `nori_regression.py` into your plugin directory (the one passed to `influxdb3 serve
+   --plugin-dir`) and use `--path nori_regression.py`.
+
+3. Set the Nori gateway key on the InfluxDB host. Get your key from the
+   [Synthefy console](https://console.synthefy.com/) (one key works for all Nori variants):
+
+   ```bash
+   export NORI_API_KEY="<your Nori API key>"
+   ```
+
+### Prerequisites
+
+- InfluxDB 3 Core or Enterprise (>= 3.8.2) with the Processing Engine enabled.
+- A Nori API key from the [Synthefy console](https://console.synthefy.com/).
+- A measurement with a numeric target field, one or more numeric feature columns, and enough labeled
+  rows (`>= min_history`) where the target is present.
+
+## Trigger setup
+
+### Scheduled trigger
+
+Every 15 minutes, fill any rows of `sensors` (for `site=A`) that are missing `pressure`, predicting
+it from `temp` and `humidity`:
+
+```bash
+influxdb3 create trigger \
+  --database mydb \
+  --path "gh:influxdata/nori_regression/nori_regression.py" \
+  --trigger-spec "every:15m" \
+  --trigger-arguments measurement=sensors,field=pressure,feature_fields="temp humidity",tags=site:A,model=synthefy/nori \
+  nori_sensors_pressure
+```
+
+### HTTP trigger
+
+```bash
+influxdb3 create trigger \
+  --database mydb \
+  --path "gh:influxdata/nori_regression/nori_regression.py" \
+  --trigger-spec "request:nori_regress" \
+  nori_http
+```
+
+## Example usage
+
+### On-demand HTTP regression
+
+Call the HTTP endpoint (exposed at `/api/v3/engine/<path>`), passing the Nori key in the header:
+
+```bash
+curl -X POST http://localhost:8181/api/v3/engine/nori_regress \
+  -H "X-Nori-Api-Key: $NORI_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"measurement":"sensors","field":"pressure","feature_fields":["temp","humidity"],"tags":{"site":"A"}}'
+```
+
+**Expected output:**
+
+```json
+{"status": "success", "task_id": "...", "result": {"status": "success", "written": 24}}
+```
+
+The predicted values are written to the `sensors_regressed` measurement.
+
+### Backfill a specific window
+
+```bash
+curl -X POST http://localhost:8181/api/v3/engine/nori_regress \
+  -H "X-Nori-Api-Key: $NORI_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"measurement":"sensors","field":"pressure","feature_fields":["temp","humidity"],"tags":{"site":"A"},"start_time":"2026-01-01T00:00:00Z","end_time":"2026-02-01T00:00:00Z"}'
+```
+
+### Dry run (preview without writing)
+
+Set `dry_run=true` to log the predictions without writing them back:
+
+```bash
+curl -X POST http://localhost:8181/api/v3/engine/nori_regress \
+  -H "X-Nori-Api-Key: $NORI_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"measurement":"sensors","field":"pressure","feature_fields":["temp","humidity"],"tags":{"site":"A"},"dry_run":true}'
+```
+
+## Output format
+
+Each prediction is written as a point:
+
+- **Measurement:** `output_measurement` (default `<measurement>_regressed`).
+- **Tags:** `model` (the slug), `source` (the input measurement), `target` (the predicted field),
+  and any single-value filter tags.
+- **Field:** `value` (float): the predicted target value.
+- **Timestamp:** the predicted row's own timestamp (nanoseconds).
+
+Example line protocol:
+
+```
+sensors_regressed,model=synthefy/nori,source=sensors,target=pressure value=1001.2 1704672000000000000
+```
+
+## Querying predictions
+
+```sql
+SELECT time, value, model, target FROM sensors_regressed ORDER BY time DESC LIMIT 24;
+```
+
+## Notes
+
+- **What it predicts:** rows in the window where the target `field` is null but every
+  `feature_fields` column is present. Rows where the target is already present become the training
+  set (`X_train`/`y_train`). If no rows are missing the target, the run is a no-op (`skipped`). It
+  does not overwrite existing target values.
+- **One series per run:** predictions are written back at each row's own timestamp, so a run must
+  resolve to a single series. If the measurement holds several series (for example one per `site`)
+  and your `tags` filter does not isolate one, two rows can share a timestamp and their output
+  points would collide, so the plugin **fails with a clear message** rather than silently dropping
+  values. Add a `tags` filter, or run one trigger per series.
+- **Features only:** Nori sees just the columns you name in `feature_fields`. Row order does not
+  matter, and no time-derived features are added.
+
+## Supported models
+
+The `model` argument is the Nori gateway slug your API key is granted:
+
+- `synthefy/nori`: the base Nori in-context tabular regression model (default).
+- `synthefy/nori-30m`: a 30M-parameter Nori variant.
+
+A single API key from the [Synthefy console](https://console.synthefy.com/) works for all of these
+variants.
+
+## Code overview
+
+### Files
+
+- `nori_regression.py`: the plugin (metadata docstring and implementation).
+- `requirements.txt`: Python dependencies.
+- `manifest.toml`: packaging metadata.
+
+### Key functions
+
+- `process_scheduled_call(influxdb3_local, call_time, args)`: scheduled entry point.
+- `process_request(influxdb3_local, query_parameters, request_headers, request_body, args)`: HTTP entry point.
+- `_query(influxdb3_local, cfg, task_id)`: reads the target and feature columns for the window into a DataFrame.
+- `_regress(influxdb3_local, cfg, api_key, task_id)`: splits labeled and null-target rows and builds `X_train`/`y_train`/`X_test`.
+- `_call_nori(...)`: sends the in-context regression request (`task="regression"`) to the gateway.
+- `_write_predictions(...)`: writes the predictions back with `write_sync`.
+
+## Troubleshooting
+
+### Common issues
+
+### Missing API key
+
+The plugin cannot find a Nori gateway key.
+
+**Solution:** set `NORI_API_KEY` on the InfluxDB host, or pass an `X-Nori-Api-Key: <key>` header
+when calling the HTTP trigger (see [Authentication](#authentication-for-the-nori-gateway)).
+
+### Model API not found (404)
+
+The gateway rejects the `model` slug for your key.
+
+**Solution:** confirm the `model` slug is spelled correctly (for example `synthefy/nori` or
+`synthefy/nori-30m`) and that you passed a valid API key from the
+[Synthefy console](https://console.synthefy.com/).
+
+### Not enough labeled rows, or nothing to predict
+
+- **`only N labeled rows (< min_history)`:** fewer than `min_history` rows have the target present.
+  Widen `window`, lower `min_history`, or check that `measurement`/`field`/`feature_fields`/`tags`
+  select data.
+- **`no rows to predict`:** every target value in the window is already present. The plugin only
+  fills rows where the target is null; there is nothing to impute.
+
+### Multiple series sharing timestamps
+
+The measurement holds more than one series and your `tags` filter did not isolate one, so
+predictions would collide on write.
+
+**Solution:** add a `tags` filter that selects a single series, or run one trigger per series.
+
+### Feature field not found
+
+A `feature_fields` column does not exist in the measurement, or it clashes with the target field or
+the reserved names `time`/`y` (not allowed).
+
+**Solution:** fix the column names in `feature_fields`.
+
+### Cold-start latency on the first call
+
+If the model has scaled to zero, the first request after idle can be slow (about 60-70 seconds) or
+return a 500 once.
+
+**Solution:** retry; the scheduled trigger recovers on the next tick. Pre-warm with a direct call
+before time-sensitive use.
+
+## Limitations
+
+- One series per run; run multiple triggers for multiple series (the plugin fails loud if a run
+  resolves to more than one series). Multi-series imputation in a single run is a possible
+  enhancement.
+- Imputes only rows where the target is null; it does not overwrite existing values.
+- Prediction quality depends on how well `feature_fields` explain the target.
+- Each call is a billed gateway request; a larger `window` sends more rows and tokens.
+
+## License
+
+Apache 2.0.
+
+## Questions/Comments
+
+Please open an issue or discussion in the
+[influxdb3_plugins](https://github.com/influxdata/influxdb3_plugins) repository.
--- a/influxdata/nori_regression/README.md
+++ b/influxdata/nori_regression/README.md
@@ -1,17 +1,20 @@
 # Nori Regression Plugin
 
-⚡ scheduled, http 🏷️ regression, tabular, machine-learning 🔧 InfluxDB 3 Core, InfluxDB 3 Enterprise
+⚡ scheduled, http 🏷️ regression, tabular, machine-learning, imputation 🔧 InfluxDB 3 Core, InfluxDB 3 Enterprise
 
-Predict a numeric field in an InfluxDB 3 measurement from other columns with **Nori**, Synthefy's
-in-context-learning tabular regression model, called through the Synthefy inference gateway. The
-plugin reads a window of rows, trains on the rows where the target field is present, predicts the
-rows where it is null (imputation / backfill), and writes the predicted values back into InfluxDB.
+> **Note:** This plugin requires InfluxDB 3.8.2 or later (it uses the synchronous write API).
+
+Predict a numeric field in an InfluxDB 3 measurement from other columns on the same rows with
+**Nori**, Synthefy's in-context-learning tabular regression model, called through the Synthefy
+inference gateway. The plugin reads a window of rows, trains on the rows where the target field is
+present, predicts the rows where it is null (imputation / backfill), and writes the predicted values
+back into InfluxDB.
 
 ## Description
 
 Nori is a tabular regression foundation model: you give it labeled feature rows (`X_train`,
 `y_train`) and query rows (`X_test`) in a single request, and it predicts a value for each query row
-in one forward pass, with no training or fine-tuning.
+in one forward pass, with no training or fine-tuning step.
 
 This plugin applies Nori to an InfluxDB measurement. You choose a target field and a set of feature
 columns; the plugin uses the rows where the target is present as the in-context training set and
@@ -30,24 +33,33 @@ Key features:
 
 - **In-context tabular regression**: no training step; the recent labeled rows are the context.
 - **Imputation / backfill**: predicts the rows where the target is null and writes them back.
-- **Scheduled or on-demand**: run on an interval (scheduled trigger) or via an HTTP endpoint.
-- **Tag filtering**: operate on a single series selected by tags.
-- **Writes results back** to InfluxDB as a new measurement using line protocol.
+- **Scheduled or on-demand**: run on an interval, or call an HTTP endpoint with an explicit window.
+- **Idempotent by default**: rows that already hold a prediction are skipped, so a repeating
+  schedule does not re-send and re-pay for the same rows.
+- **Bounded cost**: row caps and a batch size keep one run's billed rows predictable.
+- **Single-series guarantee**: a run that resolves to more than one series fails before it calls the
+  gateway, rather than training one model on two mixed series.
 
 ## Configuration
+
+Plugin parameters may be given as key-value pairs in the `--trigger-arguments` flag of
+`influxdb3 create trigger`, in the `trigger_arguments` field of the API, or entirely from a TOML
+file via `config_file_path` — see [TOML configuration](#toml-configuration). For the HTTP trigger, a
+documented subset may also be sent in the JSON request body.
 
 ### Plugin metadata
 
 This plugin includes a JSON metadata schema in its docstring that declares the supported trigger
-types (`scheduled`, `http`) and their arguments, so the InfluxDB 3 Explorer UI can render a
-configuration form.
+types (`scheduled`, `http`) and every parameter each accepts, so the
+[InfluxDB 3 Explorer](https://docs.influxdata.com/influxdb3/explorer/) UI can render a configuration
+form.
 
 ### Authentication for the Nori gateway
 
 The Nori gateway API key is a secret and is **never** read from trigger arguments or the request
 body (both are logged). It is resolved, in order:
 
-1. an incoming `X-Nori-Api-Key: <key>` request header (HTTP trigger only), then
+1. a non-empty `X-Nori-Api-Key: <key>` request header (HTTP trigger only), then
 2. the `NORI_API_KEY` environment variable set on the InfluxDB host (required for the scheduled
    trigger).
 
@@ -55,53 +67,114 @@ The key is intentionally **not** accepted in the `Authorization` header: InfluxD
 `Authorization` for its own request authorization, so a key placed there never reaches the plugin.
 Use the custom `X-Nori-Api-Key` header instead.
 
-Get a Nori API key from the [Synthefy console](https://console.synthefy.com/). A single key works
-for all Nori model variants (see [Supported models](#supported-models)), so you do not need a
-separate key per variant. This plugin does not create keys.
+Get a Nori API key from the [Synthefy console](https://console.synthefy.com/). One key covers every
+model slug its group is granted (see [Supported models](#supported-models)), so you do not normally
+need a key per variant. This plugin does not create keys.
 
-### Scheduled trigger parameters
+The gateway endpoint itself is **not** a parameter: the request carries the operator's API key and
+the training data, so a caller must never be able to choose its destination. An operator running a
+private gateway can point the plugin at it with the `NORI_GATEWAY_URL` environment variable on the
+InfluxDB host. It must be an `https://` URL; plain `http://` is accepted only for a loopback host
+(`localhost`, `127.0.0.1` or `::1`), so a local mock gateway still works in testing.
 
-| Parameter | Default | Description |
-|---|---|---|
-| `measurement` | (required) | Source measurement (table) to read from. |
-| `field` | (required) | The numeric field to predict (the regression target). The plugin trains on rows where it is present and predicts the rows where it is null. |
-| `feature_fields` | (required) | Numeric feature columns (X) used to predict `field`, **space-separated** (e.g. `temp humidity`). Use spaces, not commas (`--trigger-arguments` splits argument pairs on commas) and not dots (field names may contain a `.`). In the HTTP JSON body it may also be a list. |
-| `window` | `30d` | Time window of rows to read. Units: `s,min,h,d,w`. |
-| `tags` | (none) | Filter to a single series. Format: `key:val key2:val2` (space-separated pairs). Required if the measurement holds more than one series (see [Notes](#notes)). |
-| `model` | `synthefy/nori` | The Nori gateway slug to call. Your API key must be granted this slug. |
-| `gateway_url` | `https://inference.baseten.co/predict` | Nori gateway endpoint. |
-| `output_measurement` | `<measurement>_regressed` | Measurement to write predictions to. |
-| `target_database` | (trigger db) | Write predictions to a different database. |
-| `dry_run` | `false` | If `true`, log predictions but do not write them. |
-| `min_history` | `50` | Minimum labeled rows (target present) required to train; skip below this. |
+### Required parameters
 
-### HTTP trigger parameters
+| Parameter | Type | Default | Description |
+|---|---|---|---|
+| `measurement` | string | required | Source measurement (table) to read from. |
+| `field` | string | required | The numeric field to predict. The plugin trains on the rows where it is present and predicts the rows where it is null. |
+| `feature_fields` | string | required | Numeric feature columns (X) used to predict `field`, **space-separated** (for example `temp humidity`). Use spaces, not commas (`--trigger-arguments` splits argument pairs on commas) and not dots (a field name may contain a `.`). |
 
-All scheduled parameters are accepted (in the JSON request body or as trigger arguments), plus:
+A column name that contains a space cannot be expressed in `feature_fields` as a trigger argument,
+because every string form splits on whitespace. Name such a column from a TOML array
+(`feature_fields = ["air temp", "humidity"]`) or from a JSON list in the HTTP body.
 
-| Parameter | Default | Description |
-|---|---|---|
-| `start_time` | (none) | ISO start of the window. Overrides `window`. |
-| `end_time` | (none) | ISO end of the window. |
+### Optional parameters
 
-Scalar values in the JSON body override trigger arguments. A `tags` object may be sent as JSON
-(`{"tags": {"site": "A"}}`), and `feature_fields` may be sent as a JSON list
-(`{"feature_fields": ["temp", "humidity"]}`) or a space-separated string.
+| Parameter | Type | Default | Description |
+|---|---|---|---|
+| `window` | string | `30d` | Time window of rows to read, ending at the trigger's call time. Units: `s`, `min`, `h`, `d`, `w`, with an integer magnitude. |
+| `start_time` | string | *(none)* | ISO 8601 start of a fixed window. Given alone, the window ends now. |
+| `end_time` | string | *(none)* | ISO 8601 end of a fixed window. Given alone, the window starts one `window` earlier. |
+| `tags` | string | *(none)* | Filter to a single series. Format: `key:val key2:val2` (space-separated pairs, one value per key). A token without a `:` is rejected. Required when the window holds more than one series. |
+| `model` | string | `synthefy/nori-30m` | The Nori gateway slug to call. See [Supported models](#supported-models). |
+| `output_measurement` | string | `<measurement>_regressed` | Measurement to write predictions to. Must differ from `measurement`. |
+| `target_database` | string | *(trigger db)* | Write predictions to a different database. |
+| `dry_run` | boolean | `false` | Log the first few predictions and return them all, without writing anything. |
+| `skip_existing` | boolean | `true` | Skip rows that already hold a prediction in `output_measurement`. Set `false` to refresh earlier predictions with newer training data. |
+| `min_history` | integer | `50` | Minimum labeled rows required to train; the run is skipped below this. |
+| `max_train_rows` | integer | `1000` | Cap on labeled rows sent as the training context; the most recent rows are kept. This is the main cost control — the gateway bills per training row and column. |
+| `max_predict_rows` | integer | `5000` | Cap on rows predicted per run; the most recent rows are kept and the rest wait for a later run. |
+| `max_read_rows` | integer | `50000` | Ceiling on rows read from InfluxDB in one run, applied as a `LIMIT` on the query. The most recent rows are read, and a truncated read is logged with a warning. This bounds the plugin's memory: a row costs roughly 0.7 KB while it is held, so the default is about 35 MB. |
+| `predict_batch_size` | integer | `1000` | Rows per gateway call. Each batch re-sends the training context and is billed separately, so a larger value costs less. |
+| `request_timeout` | string | `300s` | Timeout for one gateway call. A cold start has been measured at 60-130 seconds, so keep this well above that. |
+| `max_retries` | integer | `3` | Maximum attempts per gateway call and per write. `1` disables retry. |
+| `config_file_path` | string | *(none)* | Path to a TOML file supplying every parameter, relative to `PLUGIN_DIR`. Cannot be combined with other inline arguments or a request body. |
+
+Two constraints are checked before anything runs: `min_history` must not exceed `max_train_rows`
+(no run could otherwise ever qualify), and `output_measurement` must differ from `measurement`. All
+of the integer parameters must be at least `1`.
+
+### HTTP request body parameters
+
+On the HTTP trigger, these keys may be sent in the JSON request body:
+
+`measurement`, `field`, `feature_fields`, `tags`, `window`, `start_time`, `end_time`, `dry_run`.
+
+`feature_fields` may be a JSON list (`{"feature_fields": ["temp", "humidity"]}`) or a
+space-separated string, and `tags` may be a JSON object (`{"tags": {"site": "A"}}`).
+
+**A trigger argument pins its value.** The body may fill in what the trigger left open, but it
+cannot change what the trigger already set — that is rejected. So an operator who wants the request
+to choose the measurement creates the trigger without one, and an operator who wants it fixed sets
+it as a trigger argument. This matters because `output_measurement` defaults to
+`<measurement>_regressed`: without the pin, a body-supplied `measurement` would move the write
+target too.
+
+Every other parameter is **operator-only** and is rejected by name if it appears in the body. The
+endpoint is reachable by anyone holding a database token, so the model slug (which selects a billed
+model), the write targets (`output_measurement`, `target_database`), the row caps, the timeout and
+`config_file_path` stay under the control of whoever created the trigger.
+
+`gateway_url` is not a parameter at all, in either place — use the `NORI_GATEWAY_URL` environment
+variable. Passing it (or a parameter from the plugin's earlier forecasting revision: `mode`,
+`horizon`, `step`, `lags`, `rolling`, `tz`) is rejected with a message naming the replacement, rather
+than ignored.
+
+### TOML configuration
+
+Set the `PLUGIN_DIR` environment variable and reference the file with the `config_file_path` trigger
+argument (relative paths resolve against `PLUGIN_DIR`, then `INFLUXDB3_PLUGIN_DIR`, then the parent
+of `VIRTUAL_ENV`). The TOML file then supplies **all** parameters — it is mutually exclusive with
+inline trigger arguments and with an HTTP request body. See
+[`nori_regression_config_scheduler.toml`](nori_regression_config_scheduler.toml) for an annotated
+template.
+
+```bash
+influxdb3 create trigger \
+  --database mydb \
+  --path "gh:influxdata/nori_regression/nori_regression.py" \
+  --trigger-spec "every:1h" \
+  --trigger-arguments config_file_path=nori_regression_config_scheduler.toml \
+  nori_from_toml
+```
 
 ## Requirements
 
-### Dependencies
+### Software requirements
 
-- `requests`
-- `pandas`
-- `numpy`
+- **InfluxDB 3 Core or Enterprise**, version 3.8.2 or later, with the Processing Engine enabled
+  (`influxdb3 serve --plugin-dir /path/to/plugins`).
+- **Python packages**: `influxdata-plugin-utils>=0.2.0`, `requests`.
+- A **Nori API key** from the [Synthefy console](https://console.synthefy.com/), reachable from the
+  InfluxDB host over HTTPS.
 
 ### Installation steps
 
 1. Install the Python dependencies into the InfluxDB 3 Processing Engine environment:
 
    ```bash
-   influxdb3 install package requests pandas numpy
+   influxdb3 install package influxdata-plugin-utils requests
    ```
 
 2. Reference the plugin directly from this repository with the `gh:` prefix (the form used in the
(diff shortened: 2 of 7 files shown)


influxdata/nori_forecasting/README.md
```diff
@@ -0,0 +1,363 @@
...
+This plugin includes a JSON metadata schema in its docstring that declares the supported trigger
+types (`scheduled`, `http`) and their arguments, so the InfluxDB 3 Explorer UI can render a
+configuration form.
+
+### Authentication for the Nori gateway
+
+The Nori gateway API key is a secret and is **never** read from trigger arguments or the request
+body (both are logged). It is resolved, in order:
+
+1. an incoming request header: `X-Nori-Api-Key: <key>`, or `Authorization: Api-Key <key>` (HTTP
```
Aliaksei-Kharlap: Influx uses the `Authorization` header to authorize incoming requests, so it is not possible to use it for passing other access tokens. Therefore, it is appropriate to remove any references to it—and the associated logic—from the plugin

influxdata/nori_forecasting/README.md
```diff
@@ -0,0 +1,363 @@
...
+separate key per variant. This plugin does not create keys.
+
+### Scheduled trigger parameters
+
+| Parameter | Default | Description |
+|---|---|---|
+| `measurement` | (required) | Source measurement (table) to read from. |
+| `field` | (required) | The numeric field to predict (the regression target). **forecast**: the series to forecast. **regress**: the column to predict from `feature_fields`. |
+| `mode` | `forecast` | `forecast` predicts future timestamps of `field`; `regress` predicts `field` from `feature_fields` on the same rows and fills rows where `field` is null. |
+| `feature_fields` | (none) | **regress only (required there)**: numeric feature columns (X) used to predict `field`, separated by `.` or `,`. Ignored in forecast mode. |
```
Aliaksei-Kharlap: Since Influx allows field names to contain a dot ("."), this argument format will be parsed incorrectly. There is also a parsing limitation regarding the `--trigger-arguments` parameter when using the CLI: it does not allow commas (",") within argument values, as doing so leads to incorrect parsing of the trigger arguments. A space is the preferred delimiter

influxdata/nori_forecasting/README.md
```diff
@@ -0,0 +1,363 @@
...
+| `window` | `30d` | History window read from InfluxDB. Units: `s,min,h,d,w`. |
+| `horizon` | `24` | **forecast**: number of future points to predict. |
+| `step` | `1h` | **forecast**: spacing between forecast points. Units: `s,min,h,d,w`. |
+| `lags` | `24.168` | **forecast**: lag steps (in units of `step`) used as features, separated by `.` or `,`. `--trigger-arguments` splits pairs on `,`, so use `.` there (e.g. `24.168`); commas are fine in the HTTP JSON body. Use lags `>= horizon` for a clean multi-step forecast. |
+| `rolling` | `24` | **forecast**: rolling-window size (in points) for mean/std features. `0` disables. |
+| `tags` | (none) | Filter to a single series. Format: `key:val.key2:val2`. In regress mode this is required if the measurement holds more than one series (see [Regress notes](#regress-mode-notes)). |
```
Aliaksei-Kharlap: Same remark as for the feature_fields arg

influxdata/nori_forecasting/README.md
```diff
@@ -0,0 +1,363 @@
...
+| `tz` | `UTC` | **forecast**: IANA timezone for calendar features (e.g. `America/New_York`) so hour-of-day and day-of-week align to the series' local seasonality. |
+| `model` | `synthefy/nori` | The Nori gateway slug to call. Your API key must be granted this slug. |
+| `gateway_url` | `https://inference.baseten.co/predict` | Nori gateway endpoint. |
+| `output_measurement` | (per mode) | Measurement to write predictions to. Default `<measurement>_forecast` (forecast) or `<measurement>_regressed` (regress). |
+| `target_database` | (trigger db) | Write predictions to a different database. |
+| `dry_run` | `false` | If `true`, log predictions but do not write them. |
+| `min_history` | `50` | Minimum labeled/context rows required; skip below this. |
+| `config_file_path` | (none) | TOML file (under `PLUGIN_DIR`) whose values override arguments. |
```
Aliaksei-Kharlap: The documentation describes this parameter and the option to use a TOML configuration, but the plugin code lacks this logic

influxdata/nori_forecasting/nori_forecasting.py
```diff
@@ -0,0 +1,535 @@
...
+    step = _interval_to_timedelta(cfg["step"])
+    for lag in cfg["lags"]:
+        lagged_times = times - lag * step
+        # ffill (not "nearest"): a lag uses the most recent value AT OR BEFORE the lagged time,
+        # never a later one, so the feature can't leak a future value; tolerance bounds the reach.
+        feats[f"lag_{lag}"] = series.reindex(
+            lagged_times, method="ffill", tolerance=step
+        ).to_numpy()
+    if cfg["rolling"] > 0 and len(series) >= cfg["rolling"]:
+        roll_mean = series.rolling(cfg["rolling"]).mean()
```
Aliaksei-Kharlap: eries.rolling(N).mean()/.std() includes the current point, so the roll_mean/roll_std for row t already contain y[t] — the value
  we're trying to predict. This leaks the label into training (and contradicts the docstring)

Aliaksei-Kharlap (changes requested):

ajanbekzat:
@Aliaksei-Kharlap thanks for the review, and for the detailed comments. One larger change first: we have refocused this plugin to be **regression-only** and renamed it (`nori_forecasting` to `nori_regression`). We removed the time-series forecast mode to keep the plugin focused on Nori's core capability, tabular regression. The plugin now predicts a target field from other columns on the same rows and fills (imputes) the rows where the target is null.

How each comment is addressed:

1. **Rolling-window label leakage** (`rolling(N).mean()` includes the current point). Resolved: that feature-engineering code was removed along with the forecast mode.
2. **`Authorization` header.** Removed. The key is no longer read from the incoming `Authorization` header (InfluxDB uses that header for its own request authorization, so it never reached the plugin anyway); it is resolved only from an `X-Nori-Api-Key` header or the `NORI_API_KEY` environment variable. Both the references and the logic are gone.
3. **`feature_fields` delimiter.** Now space-separated (for example `temp humidity`). It no longer splits on `.` (InfluxDB field names can contain a dot) or on `,` (which `--trigger-arguments` uses to split argument pairs). A JSON list is also accepted in the HTTP request body.
4. **`tags` delimiter.** Same fix: space-separated `key:val` pairs.
5. **`config_file_path` / TOML configuration.** Removed from the metadata and README, since the plugin did not implement it.

CI is green. This is ready for another review, thank you.

Aliaksei-Kharlap:
Hi, I'll check out the plugin soon, thanks!

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    quote-escaped, so quotes in values/names can't break or inject the query. Returns
+    (time_clause, tag_clause, params).
+    """
+    params: dict = {}
+    tag_clause = ""
+    for i, (k, v) in enumerate(cfg["tags"].items()):
+        p = f"tag{i}"
+        tag_clause += f" AND {_ident(k)} = ${p}"
+        params[p] = v
+    if cfg["start_time"] and cfg["end_time"]:
```
Aliaksei-Kharlap: The docstring states that `end_time` is optional and `start_time` can be used independently, but here `start_time` is ignored unless `end_time` is specified

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    """Resolve the gateway key: the incoming `X-Nori-Api-Key` header wins (HTTP trigger), else the
+    NORI_API_KEY env var.
+
+    The key is deliberately NOT read from the incoming `Authorization` header: InfluxDB parses that
+    header for its own request authorization, so a custom scheme there does not reach the plugin.
+    """
+    if request_headers:
+        for k, v in request_headers.items():
+            if k.lower() == "x-nori-api-key":  # case-insensitive per server
+                return v.strip()
```
Aliaksei-Kharlap: I think it's worth adding a check for an empty string here

Aliaksei-Kharlap (changes requested):

caterryan (changes requested):
Superseded. I re-posted these findings as line-level comments on the code they refer to. See my newer review.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+def process_request(influxdb3_local, query_parameters, request_headers, request_body, args=None):
+    task_id = str(uuid.uuid4())
+    try:
+        body = {}
+        if request_body:
+            body = json.loads(request_body)
+        # scalar body values override trigger args
+        merged = {**(args or {}), **{k: str(v) for k, v in body.items() if not isinstance(v, (list, dict))}}
```
caterryan: **Blocker: a caller can redirect the operator's API key to any host.**

This line merges every scalar body key over the trigger arguments. `_build_config:126` then reads `gateway_url` from the merged map, and `_call_nori:232` posts to that URL with `Authorization: Api-Key <NORI_API_KEY>`.

A caller who holds only a database token therefore obtains three things:
- the operator's `NORI_API_KEY`, a credential that the operator never shared;
- the `X_train` and `y_train` data;
- a request that starts inside the InfluxDB host and goes to any URL.

The same merge also exposes `target_database`, `model`, `min_history`, `window`, `output_measurement`, and `dry_run`. A caller can write to another database, select a more expensive model slug, or set `min_history` to 0.

Do not patch one key. Define an explicit list of body-overridable keys and drop every other key. Keep `gateway_url` and `target_database` out of that list. Add a regression test, because a security fix with no test returns.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    return {
+        "measurement": measurement,
+        "field": field,
+        # feature columns (X); space-separated (see _split_list). dict.fromkeys de-duplicates while
+        # preserving order (a repeated column would otherwise produce a duplicate SQL projection).
+        "feature_fields": list(dict.fromkeys(_split_list(args.get("feature_fields", "")))),
+        "tags": tags,
+        "window": args.get("window", "30d"),
+        "model": args.get("model", DEFAULT_MODEL_SLUG),
+        "gateway_url": args.get("gateway_url", DEFAULT_GATEWAY_URL),
```
caterryan: **Blocker (see line 380).** `gateway_url` comes from the arguments, and `process_request:380` merges the request body over those arguments. A caller therefore chooses the host that receives the `Authorization: Api-Key` header at line 235.

Make the URL a module constant, as `synthefy_forecasting.py:78` does with `SYNTHEFY_API_BASE_URL`. Apply the same rule to `target_database` at line 128.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+        f"n_features={n_features} n_train={len(X_train)} n_test={len(X_test)}"
+    )
+    payload = {"model": cfg["model"], "task": "regression",
+               "X_train": X_train, "y_train": y_train, "X_test": X_test}
+    resp = requests.post(
+        cfg["gateway_url"],
+        json=payload,
+        headers={"Content-Type": "application/json", "Authorization": f"Api-Key {api_key}"},
+        timeout=120,
+    )
```
caterryan: **Blocker (see line 380).** This call sends the operator's `NORI_API_KEY` to `cfg["gateway_url"]`, and a caller controls that value. The `X_train` and `y_train` data go with it. Require HTTPS.

**Second defect: no retry.** This call fails after one attempt, but `_write_predictions:330` retries three times with backoff. The README documents a cold-start 500 and a 60-70 second latency at lines 297-303. The retry effort therefore sits on the reliable side of the system. Retry a 5xx response, a 429 response, and a timeout. Make the 120-second timeout configurable.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    """Parse trigger args (strings) into a typed config with defaults."""
+    args = args or {}
+    measurement = args.get("measurement")
+    field = args.get("field")
+    tags = {}
+    if args.get("tags"):
+        for pair in args["tags"].split():  # space-separated key:val pairs (a value may contain '.')
+            if ":" in pair:
+                k, v = pair.split(":", 1)
+                tags[k.strip()] = v.strip()
```
caterryan: A token that holds no colon is discarded without a message. `tags=site=A` therefore gives an empty filter, and the query reads every series in the measurement.

The failure is silent, and it widens the query. Raise a `ValueError` for each invalid token.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+        if isinstance(body.get("tags"), dict):
+            cfg["tags"] = {str(k): str(v) for k, v in body["tags"].items()}
+        if isinstance(body.get("feature_fields"), list):
+            cfg["feature_fields"] = list(dict.fromkeys(str(x) for x in body["feature_fields"]))
+        api_key = _get_api_key(request_headers)
+        result = _run(influxdb3_local, cfg, api_key, task_id)
+        # surface the real outcome at the top level (success / skipped / dry_run), not always
+        # "success", so a caller can tell a real prediction from a no-op.
+        return {"status": result.get("status", "success"), "task_id": task_id, "result": result}
+    except json.JSONDecodeError:
```
caterryan: `requests.exceptions.JSONDecodeError` is a subclass of `json.JSONDecodeError` (verified, requests 2.32.5). A failure of `resp.json()` at line 239 therefore lands in this handler.

A gateway that returns HTTP 200 with a non-JSON body gives the caller `invalid JSON in request body`, for a request whose body was valid JSON. The operator then debugs the wrong system.

Parse `request_body` in its own `try`/`except` block at line 378.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    resp = requests.post(
+        cfg["gateway_url"],
+        json=payload,
+        headers={"Content-Type": "application/json", "Authorization": f"Api-Key {api_key}"},
+        timeout=120,
+    )
+    resp.raise_for_status()
+    result = resp.json()
+    preds = result.get("predictions")
+    if not isinstance(preds, list) or len(preds) != len(X_test):
```
caterryan: The check tests the type and the length of the list, but not the elements. Two failures pass it.

- **A list of nulls passes.** `_write_predictions:315` skips each null and returns 0, and `_run:358` reports `{"status":"success","written":0}`. A complete upstream failure is then indistinguishable from a healthy no-op.
- **A `float("nan")` element passes.** It reaches `LineBuilder.float64_field` at line 323, and the write fails with a line-protocol error that never names the gateway.

Require a finite real number for every element here, at the gateway boundary. Note that a numeric string coerces without an error at line 323, so a type test is necessary as well as a finite test.

influxdata/nori_regression/nori_regression.py
caterryan: Two defects on this line.

1. **An empty header value wins over the environment variable.** `X-Nori-Api-Key: ` returns an empty string, the `NORI_API_KEY` fallback never runs, and the plugin sends `Authorization: Api-Key ` with no key. Test the value for truth before you return it.
2. **`.strip()` assumes a string.** A list value raises `AttributeError`. Add an `isinstance(v, str)` test.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    cols = {r.get("column_name") for r in (rows or [])}
+    if not cols:
+        raise ValueError(f"measurement {cfg['measurement']!r} not found (no columns)")
+    if cfg["field"] not in cols:
+        raise ValueError(f"target field {cfg['field']!r} not found in {cfg['measurement']!r}")
+    missing = [k for k in cfg["tags"] if k not in cols]
+    if missing:
+        raise ValueError(f"tag column(s) {missing} not found in {cfg['measurement']!r}")
+    if not cfg["feature_fields"]:
+        raise ValueError("`feature_fields` is required (the '.'/','-separated feature columns X)")
```
caterryan: The message says `'.'/','-separated`. The docstring at line 99 and the README both say space-separated, and both forbid dots, because a field name may contain a dot. Correct the message.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+        {"name": "dry_run", "example": "false", "description": "If true, log predictions but do not write them.", "required": false},
+        {"name": "min_history", "example": "50", "description": "Minimum labeled rows (target present) required to train; abort below this.", "required": false}
+    ],
+    "http_args_config": [
+        {"name": "measurement", "example": "sensors", "description": "Source measurement. May also be provided in the JSON request body.", "required": false},
+        {"name": "field", "example": "pressure", "description": "Target field to predict. May also be in the request body.", "required": false},
+        {"name": "feature_fields", "example": "temp humidity", "description": "Feature columns. In the JSON request body this may be a JSON list of column names, or a space-separated string.", "required": false},
+        {"name": "start_time", "example": "2026-01-01T00:00:00Z", "description": "Optional ISO start of the window (overrides `window`).", "required": false},
+        {"name": "end_time", "example": "2026-02-01T00:00:00Z", "description": "Optional ISO end of the window.", "required": false}
+    ]
```
caterryan, line 47: `http_args_config` declares 5 parameters. `_build_config` accepts 13.

Explorer will not render `tags`, `window`, `model`, `output_measurement`, `target_database`, `dry_run`, `min_history`, or `gateway_url`, and a user of the UI cannot set them. Declare every parameter that the HTTP path accepts.

influxdata/nori_regression/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+If the model has scaled to zero, the first request after idle can be slow (about 60-70 seconds) or
+return a 500 once.
+
+**Solution:** retry; the scheduled trigger recovers on the next tick. Pre-warm with a direct call
+before time-sensitive use.
+
+## Limitations
+
+- One series per run; run multiple triggers for multiple series (the plugin fails loud if a run
```
caterryan: This statement is not true.

`_regress:273` tests for duplicate timestamps inside the test set only. Two series that hold different timestamps pass the guard, train as one series, and write output points that carry no source tag.

Correct the sentence, or make the code meet it. See my comment at `nori_regression.py:273`.

influxdata/library/plugin_library.json
```diff
@@ -307,6 +307,17 @@
...
             },
+            {
+                "name": "Nori Regression",
+                "path": "influxdata/nori_regression/nori_regression.py",
+                "description": "[BETA] Requires InfluxDB 3.8.2+. Predict a numeric field in an InfluxDB 3 measurement from other columns with Nori, Synthefy's in-context-learning tabular regression model, via the Baseten gateway. Trains on the rows where the target field is present and predicts (imputes) the rows where it is null, writing the predictions back to InfluxDB. Scheduled and HTTP triggers; the model gateway slug is configurable (synthefy/nori or synthefy/nori-30m).",
+                "author": "Synthefy",
+                "docs_file_link": "https://github.com/influxdata/influxdb3_plugins/blob/main/influxdata/nori_regression/README.md",
+                "required_plugins": [],
+                "required_libraries": ["requests", "pandas", "numpy"],
+                "last_update": "2026-07-15",
```
caterryan: `REQUIRED_PLUGIN_METADATA.md:55` states that this file is generated from the plugin docstring metadata by the repository owners, and that it must not be edited by hand. Remove this entry.

The date is also stale. The branch changed after 2026-07-15.

influxdata/nori_regression/README.md
```diff
@@ -0,0 +1,321 @@
...
+| `feature_fields` | (required) | Numeric feature columns (X) used to predict `field`, **space-separated** (e.g. `temp humidity`). Use spaces, not commas (`--trigger-arguments` splits argument pairs on commas) and not dots (field names may contain a `.`). In the HTTP JSON body it may also be a list. |
+| `window` | `30d` | Time window of rows to read. Units: `s,min,h,d,w`. |
+| `tags` | (none) | Filter to a single series. Format: `key:val key2:val2` (space-separated pairs). Required if the measurement holds more than one series (see [Notes](#notes)). |
+| `model` | `synthefy/nori` | The Nori gateway slug to call. Your API key must be granted this slug. |
+| `gateway_url` | `https://inference.baseten.co/predict` | Nori gateway endpoint. |
+| `output_measurement` | `<measurement>_regressed` | Measurement to write predictions to. |
+| `target_database` | (trigger db) | Write predictions to a different database. |
+| `dry_run` | `false` | If `true`, log predictions but do not write them. |
+| `min_history` | `50` | Minimum labeled rows (target present) required to train; skip below this. |
```
caterryan: `scripts/validate_readme.py --plugins nori_regression` gives 0 errors and **7 warnings**. Five sections are absent: `### Optional parameters`, `### TOML configuration`, `### Data requirements`, `### Software requirements`, and `### Schema requirements`. An `influxdb3 query` example and an `influxdb3 write` example are also absent.

The PR description reports `0 errors` and omits the warnings.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    # (a multi-series measurement not isolated by `tags`), their output points would have an
+    # identical measurement+tag+time key and silently overwrite each other, so fail loud instead.
+    dup = pd.Index(test["time"])[pd.Index(test["time"]).duplicated()].unique()
+    if len(dup):
+        raise ValueError(
+            f"matched multiple series sharing {len(dup)} timestamp(s), so writing predictions "
+            f"back would collide and silently drop values. Add a `tags` filter that isolates a "
+            f"single series (or run one trigger per series). First colliding time: "
+            f"{pd.Timestamp(dup[0])}"
+        )
```
caterryan: This guard tests for duplicate timestamps inside the **test set only**. Two series that hold different timestamps pass it.

The result: `X_train` holds the rows of both series in one training context, and `_write_predictions:321` gives the output points no source tag, because `cfg["tags"]` is empty. The user gets one model that mixes two series, and an output that cannot be traced back to either.

`README.md:307` states that the plugin fails loud here. It does not.

Count the distinct tag combinations in the window and fail above one. As an alternative, carry the source tags onto the output points and support several series in one run.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    """
+    feats = cfg["feature_fields"]
+    time_clause, tag_clause, params = _where_clause(cfg)
+    feat_cols = ", ".join(_ident(f) for f in feats)
+    sql = (
+        f"SELECT time, {_ident(cfg['field'])} AS y, {feat_cols} "
+        f"FROM {_ident(cfg['measurement'])} "
+        f"WHERE {time_clause}{tag_clause} "
+        f"ORDER BY time"
+    )
```
caterryan, line 550: The query has no `LIMIT`, and the payload has no cap or batch logic. A 30-day window at 1 Hz is about 2.6 M rows in one POST, against the fixed 120-second timeout at line 237. `min_history` guards the floor only.

Add `max_train_rows` and `max_predict_rows`. Keep the most recent rows. Send `X_test` in batches.

influxdata/nori_regression/nori_regression.py
caterryan: This test needs both bounds. `start_time` alone, and `end_time` alone, both fall to the `INTERVAL '30d'` branch with no message, and the call still returns success. A user who asks for one bound silently gets a different window.

Apply each bound on its own, or reject an incomplete range.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    if not cfg["feature_fields"]:
+        raise ValueError("`feature_fields` is required (the '.'/','-separated feature columns X)")
+    reserved = [f for f in cfg["feature_fields"] if f in ("time", "y") or f == cfg["field"]]
+    if reserved:
+        raise ValueError(
+            f"feature_fields cannot include the target field or the reserved names time/y: {reserved}"
+        )
+    missing_f = [f for f in cfg["feature_fields"] if f not in cols]
+    if missing_f:
+        raise ValueError(f"feature field(s) {missing_f} not found in {cfg['measurement']!r}")
```
caterryan: `_validate_columns` tests the column name only, not the data type.

A tag column passes here. `_query:215` then coerces every value to null, and `_regress:256` drops every row. The user sees `only 0 labeled rows (< min_history 50)`, which never names the real fault.

Read `data_type` from `information_schema.columns` and reject a non-numeric feature column here.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+                raise
+    return 0
+
+
+def _run(influxdb3_local, cfg, api_key, task_id) -> dict:
+    if not cfg["measurement"] or not cfg["field"]:
+        raise ValueError("`measurement` and `field` are required")
+    _validate_columns(influxdb3_local, cfg)
+    out_times, preds = _regress(influxdb3_local, cfg, api_key, task_id)
+    if not preds:
```
caterryan, line 1008: A list of nulls is truthy, so this test passes, and the run reports `success` with `written: 0`. Validate the predictions at line 241 instead.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+# --- Entry points ----------------------------------------------------------
+
+def process_scheduled_call(influxdb3_local, call_time, args=None):
+    task_id = str(uuid.uuid4())
+    try:
+        cfg = _build_config(args)
+        api_key = _get_api_key()  # scheduled: no request headers -> env var only
+        _run(influxdb3_local, cfg, api_key, task_id)
+    except Exception as e:  # noqa: BLE001 - never let a scheduled run crash the engine
+        influxdb3_local.error(f"[{task_id}] {type(e).__name__}: {e}")
```
caterryan: **Follow-up, not a merge gate.** Each scheduled run reads the same window, and the predictions go to `<measurement>_regressed`, so the source rows stay null and nothing converges. The next run therefore sends a byte-identical payload to a metered endpoint. At `every:15m` with `window=30d`, the plugin sends each row about 2,880 times.

Add a `skip_existing` flag with a default of true: read `output_measurement` first and remove the timestamps that already hold a prediction.

Keep the flag rather than remove the behaviour, because later training data can improve an earlier prediction.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+def _split_list(text) -> list:
+    """Parse a space- (or comma-) separated arg into a list of trimmed, non-empty strings.
+
+    NOT '.'-separated: InfluxDB field names may contain dots. Prefer spaces in `--trigger-arguments`
+    (the CLI splits argument pairs on commas, so a comma inside a value breaks parsing); commas are
+    fine inside the HTTP JSON body.
+    """
+    return [x for x in re.split(r"[,\s]+", str(text).strip()) if x]
```
caterryan: **Follow-up.** Every supported string format splits on whitespace, so a field name that contains a space is unreachable from a scheduled trigger. `feature_fields=air temp` gives `["air", "temp"]`, and then `feature field(s) not found`.

The JSON list form at line 386 works over HTTP. Document the limit for the scheduled trigger.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+        result = _run(influxdb3_local, cfg, api_key, task_id)
+        # surface the real outcome at the top level (success / skipped / dry_run), not always
+        # "success", so a caller can tell a real prediction from a no-op.
+        return {"status": result.get("status", "success"), "task_id": task_id, "result": result}
+    except json.JSONDecodeError:
+        influxdb3_local.error(f"[{task_id}] invalid JSON body")
+        return {"status": "failed", "message": "invalid JSON in request body"}
+    except Exception as e:  # noqa: BLE001
+        influxdb3_local.error(f"[{task_id}] {type(e).__name__}: {e}")
+        return {"status": "failed", "message": str(e)}
```
caterryan: **Follow-up.** The response returns `str(e)` to the caller. That text can hold the full gateway URL, the target database name, and the storage error detail.

Return a stable public message and write the detail to the log. This pattern is repository-wide, so a repository-wide fix is better than a fix here alone.

influxdata/nori_regression/nori_regression.py
```diff
@@ -0,0 +1,397 @@
...
+    batch = _BatchLines(builders)
+    db = cfg["target_database"]
+    for attempt in range(max_retries):
+        try:
+            if db:
+                influxdb3_local.write_sync_to_db(db, batch, no_sync=True)
+            else:
+                influxdb3_local.write_sync(batch, no_sync=True)
+            return len(builders)
+        except Exception as e:  # noqa: BLE001
```
caterryan: **Follow-up.** The retry catches every exception. A permanent fault, such as a schema conflict, costs about 5 seconds and then fails anyway. Narrow the retry to a transport error and a timeout.

caterryan (changes requested):
Had a few agents do reviews and manual validations, and they found the following issues that might need to be addressed.
I wouldn't consider any of these "required", so if you don't agree, that's fine, we can ignore.


## 1. Blocker: a caller can redirect the operator's API key

`process_request` merges every scalar body key over the trigger arguments, and `gateway_url` is one of those keys. A caller who holds only a database token therefore obtains the operator's `NORI_API_KEY`, the `X_train` and `y_train` data, and a request that starts inside the InfluxDB host and goes to any URL. The same merge also exposes `target_database`, `model`, and `min_history`.

**Fix:** define an explicit list of body-overridable keys and drop every other key. Keep `gateway_url` and `target_database` out of that list. Add a regression test.

## 2. Fix before merge

Each item below is small.

| Location | Finding |
|---|---|
| [`:116`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532046) | A `tags` token with no colon is discarded in silence, so the query reads every series |
| [`:392`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532049) | A gateway fault reports `invalid JSON in request body`, which blames the caller |
| [`:241`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532053) | Predictions are not validated: a list of nulls reports success, and a `NaN` fails at the storage layer |
| [`:351`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532126) | The same defect, seen from the caller side |
| [`:86`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532061) | An empty `X-Nori-Api-Key` value suppresses the environment fallback |
| [`:155`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532066) | The error text contradicts the documented format |
| [`:23`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532071) | `http_args_config` declares 5 parameters. `_build_config` accepts 13. |
| [`plugin_library.json:318`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532088) | `REQUIRED_PLUGIN_METADATA.md:55` forbids a hand edit of this file |

## 3. The documents make three claims that the code does not meet

| Location | Claim |
|---|---|
| [`:280`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532104) and [`README.md:307`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532077) | "the plugin fails loud if a run resolves to more than one series". The guard tests the test set for duplicate timestamps only, so two series that hold different timestamps train as one. |
| [`README.md:76`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532096) | The PR description reports "0 errors". `scripts/validate_readme.py` also gives 7 warnings. |
| PR description | `config_file_path` is listed under "Configuration highlights". The code has no such support, and passing the value changes nothing. 28 of 33 plugins accept it. Delete the claim, or add the feature. |

## 4. Recommended, but not a merge gate

| Location | Finding |
|---|---|
| [`:205`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532109) | The query has no `LIMIT`. A 30-day window at 1 Hz is about 2.6 M rows in one POST. |
| [`:237`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532040) | The gateway call has no retry, but the local write retries three times. The README documents a cold-start 500. |
| [`:179`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532114) | `start_time` alone, and `end_time` alone, both fall back to the 30-day window in silence |
| [`:163`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532119) | A non-numeric feature column passes validation, and the user then sees a `min_history` message |

**No test file exists.** `_interval_to_timedelta`, `_split_list`, `_build_config`, `_where_clause`, and `_get_api_key` are pure and need no mock server. Four of the comments above live in them.

## 5. Other Findings

| Location | Finding |
|---|---|
| [`:370`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532129) | Each scheduled run sends a byte-identical payload to a metered endpoint. At `every:15m` with `window=30d`, that is about 2,880 sends per row. The fix needs a flag, so design it rather than rush it. |
| [`:103`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532133) | A field name that contains a space is unreachable from a scheduled trigger |
| [`:397`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532136) | The response returns `str(e)`, which can hold the gateway URL and the database name. The pattern is repository-wide. |
| [`:337`](https://github.com/influxdata/influxdb3_plugins/pull/115#discussion_r3675532139) | The write retry catches every exception, so a permanent fault costs about 5 seconds |

---

ajanbekzat:
@caterryan @Aliaksei-Kharlap thank you both — this was a genuinely useful review. Every item is addressed below, and I took the blocker seriously enough to remove the parameter it turned on rather than patch the merge.

**Before the point-by-point: your blocker had a second half, and it reaches beyond this PR.** Fixing the key-name merge was necessary but not sufficient — the allowlist gates key *names*, while the *values* still flowed into `influxdata_plugin_utils.config.load_plugin_config`, and dynaconf evaluates converter tokens on any string starting with `@`. So this, from a caller holding only a database token:

```bash
curl -X POST .../api/v3/engine/nori_regress -d '{"measurement":"@format {env[NORI_API_KEY]}", ...}'
# {"status":"failed","message":"target field 'p' not found in 'SECRET-KEY-abc123'. ..."}
```

returned the operator's `NORI_API_KEY` in the error message. `@read_file /path` returned arbitrary host file contents the same way, which also reinstates the arbitrary-file-read that keeping `config_file_path` out of the body was meant to prevent. I've fixed it here by refusing any argument or body value that begins with `@` (all ~30 dynaconf tokens start with that one character, and nothing else triggers them, so the guard is complete; `a@b` still works).

**This is not specific to my plugin.** The shared helper resolves the token, so any plugin that passes an HTTP body through `load_plugin_config` has it. `gapfill` on `main` does exactly that (`process_request` → `load_plugin_config(data, …)`), and I reproduced the leak through the helper directly:

```
load_plugin_config({"source_measurement": "@format {env[SECRET_ENV]}"}, ..., source="args")
  -> cfg.source_measurement == 'LEAKED-abc123'
load_plugin_config({"source_measurement": "@read_file /tmp/planted2.txt"}, ...)
  -> cfg.source_measurement == 'FILE-abc123'
```

The right fix is in `influxdata-plugin-utils` — either refuse tokens on the untrusted layer or construct `Dynaconf` so they are never evaluated — so that every plugin gets it rather than each one remembering. Happy to open that PR separately; tell me if you'd rather own it. Worth a look at the other `process_request` implementations either way.

Two more things your agents could not have caught, which I found while re-testing against the live gateway:

1. **The default `model` slug was broken.** `synthefy/nori` returns `404 {"error":"please check the model you provided"}` — Synthefy retired the bare slug in favour of size-explicit ones. The plugin's default configuration could never have worked. The default is now `synthefy/nori-30m`, `synthefy/nori-6m` is documented as the cheaper option, and the retired slug is rejected with a pointed message instead of a 404 from the gateway.
2. **The fixed 120-second timeout was below the cold-start time.** Measured cold starts: 69 s for `nori-6m`, **125.5 s** for `nori-30m`. The old timeout would have failed every cold start on the larger model. `request_timeout` is now configurable and defaults to `300s`, matching Synthefy's own client.

## 1. Blocker: a caller can redirect the operator's API key

Fixed, and wider than asked.

- **`gateway_url` is no longer a parameter at all.** It is a module constant, with an operator-only `NORI_GATEWAY_URL` environment override for private gateways — the same trust level as `NORI_API_KEY`, and not logged with the trigger arguments. The override must be `https://` (plain `http://` only for loopback), parsed with `urlsplit` so `http://localhost@evil.example/` resolves to `evil.example` and is rejected.
- **An explicit `BODY_OVERRIDABLE_KEYS` allowlist**: `measurement`, `field`, `feature_fields`, `tags`, `window`, `start_time`, `end_time`, `dry_run`. Everything else — `target_database`, `output_measurement`, `model`, `min_history`, the row caps, `request_timeout`, `max_retries`, `skip_existing`, `config_file_path` — is operator-only.
- **A trigger argument now pins its value.** The allowlist alone was not enough: `output_measurement` defaults to `<measurement>_regressed`, and `measurement` is body-settable, so a body could still re-point the write target. The body may now fill in only what the trigger left open; overriding a value the operator set is rejected. An operator who wants the caller to choose creates the trigger without that argument.
- `gateway_url` as a *trigger argument* is rejected too, rather than silently ignored — as are the parameters from the earlier forecasting revision (`mode`, `horizon`, `step`, `lags`, `rolling`, `tz`), each with a message naming the replacement.
- Rejected keys are **named in the error**, not dropped silently, because a silent drop is the same class of bug as the rest of your review:

  ```
  request body may not set ['gateway_url']: these are operator settings. Set them as
  trigger arguments (or in the TOML config file) instead. Body-overridable keys:
  ['dry_run', 'end_time', 'feature_fields', 'field', 'measurement', 'start_time', 'tags', 'window'].
  ```

- Regression tests: one parametrized case per operator key (12), plus a test asserting `gateway_url` is absent from the config even as a *trigger* argument. Verified live against 3.10.3.

## 2. Fix before merge

| Item | Resolution |
|---|---|
| `:116` tags token with no colon | Now uses `influxdata_plugin_utils.parsing.parse_key_value(pair_sep=" ", kv_sep=":")`, which raises. `tags=site=A` → `invalid `tags`: Invalid pair 'site=A' (missing ':')`. |
| `:392` gateway fault blamed the caller | Body decoding moved into its own `_decode_body` with its own `try`/`except`. A gateway that answers 200 with HTML now raises `GatewayError("gateway returned a non-JSON body (HTTP 200, content-type 'text/html')")`. Your subclass observation was exactly right, and this is a real cold-start behaviour, not a hypothetical. |
| `:241` predictions not validated | Validated element-wise at the gateway boundary, with one nuance: the gateway **legitimately** emits JSON `null` for a row whose prediction is not finite (Starlette runs `allow_nan=False`, so null is how non-finite is expressed). So a null is a per-row outcome — skipped, warned and counted — while a **non-null** non-numeric or non-finite element fails the batch, and an **all-null** batch fails. That closes the "success with `written: 0`" hole without failing a whole batch over one bad row. If you would rather any null fail the batch, say so and I will flip it. |
| `:351` same defect, caller side | Closed by the above. `_run` no longer treats a truthy list as proof of work, and it now distinguishes a partial run: a gateway fault part-way through returns `{"status":"partial","written":8,"remaining":12}` rather than `success`. The batches that already returned are still written, because they were already billed. |
| `:86` empty `X-Nori-Api-Key` | An empty or whitespace value now falls through to the environment. Also handles a list value (repeated header) and a non-string key without `AttributeError`. |
| `:155` error text contradicted the docs | Corrected to "the space-separated numeric feature columns (X)". |
| `:23` `http_args_config` declared 5 of 13 | Both `scheduled_args_config` and `http_args_config` now declare all **19** parameters, and the descriptions mark which are trigger-argument-only. A test asserts the declared set equals the accepted set, so this cannot drift again. |
| `plugin_library.json:318` | **Respectfully pushing back — see below.** |

### On `plugin_library.json`

`REQUIRED_PLUGIN_METADATA.md:55` does say the file is generated and should not be hand-edited, but current practice on `main` is the opposite: every recent plugin PR edited it in the plugin's own commit.

```
$ git show --stat 9f499a7   # gapfill (#128)
 influxdata/library/plugin_library.json | 11 +
$ git show --stat fd06cd5   # resampler (#125)
 influxdata/library/plugin_library.json | 10 +
$ git show --stat 3bd4bbc   # signal_filter (#121)
 influxdata/library/plugin_library.json | 13 +-
```

So I kept the entry and matched that convention instead: fixed the stale date (now `2026-07-30`), mirrored `required_libraries` from `manifest.toml` verbatim, and dropped the retired slug from the description. Happy to remove it if you would rather the doc win — but then `REQUIRED_PLUGIN_METADATA.md` and the last three merged plugin PRs disagree, which is worth settling separately.

## 3. Documented claims the code did not meet

All three now hold.

- **"fails loud if a run resolves to more than one series."** You were right that it did not. The guard now counts **distinct tag combinations** in the window, not duplicate timestamps, so two series with disjoint timestamps are caught. Verified live with exactly that shape:

  ```
  the window holds 2 series (e.g. {site=A}, {site=B}), but a run must resolve to one:
  predictions are written at each row's own timestamp, so several series would train as
  one model and collide on write. Add a `tags` filter that isolates a single series...
  ```

  I also took your alternative in part: output points now carry **every tag of the source series**, not only the tags the caller filtered on, so a prediction is traceable even when no `tags` filter was given.
- **README warnings.** Fair hit — the description reported errors and omitted warnings. It is now **0 errors and 0 warnings**; all five missing sections and both missing command examples were added.
- **`config_file_path`.** Implemented rather than deleted, via `influxdata_plugin_utils.config.load_plugin_config`, the same call shape gapfill and resampler use — TOML supplies every parameter and is mutually exclusive with inline arguments and with a request body. An annotated `nori_regression_config_scheduler.toml` ships with the plugin. A TOML array also turns out to be the only way to name a feature column containing a space, which resolves your `:103` note too.

## 4. Recommended

| Item | Resolution |
|---|---|
| `:205` no `LIMIT` | Two parts. The **payload** is bounded by `max_train_rows` (1000), `max_predict_rows` (5000) and `predict_batch_size` (1000), keeping the most recent rows and logging what was dropped. The **read** is bounded by an actual `ORDER BY time DESC LIMIT` — you were right that caps applied after materialisation are not a fix. I measured your scenario before adding it: a 30-day window at 1 Hz peaked at **1.8 GB** in the plugin to send 10,000 rows. `max_read_rows` (50000, operator-only) now caps that at roughly 35 MB, and a truncated read warns rather than passing silently. One caveat worth recording: **batching is not free** — each batch re-sends `X_train`, and billing is per request on `X_train` rows × columns, so splitting `X_test` into N batches bills N times. The log says so, and the README has a "Cost and metering" section. |
| `:237` no retry on the gateway call | Retries 408/409/425/429 and all 5xx plus connection errors, honours `Retry-After`, exponential backoff with jitter capped at 30 s. Read timeouts are deliberately **not** retried — they have already spent the whole budget, and per Synthefy's own client a timeout usually means the key's group was never granted the slug. Timeout configurable, default `300s`. |
| `:179` one bound silently ignored | Each bound now applies on its own: `start_time` alone reads up to now, `end_time` alone reads one `window` before it. The `INTERVAL` string interpolation is gone entirely — both bounds are computed in Python and passed as bound parameters. |
| `:163` non-numeric feature column | `_resolve_schema` reads `data_type` from `information_schema.columns` and rejects a non-numeric target or feature up front: `feature field 'site' is Dictionary(Int32, Utf8), not a numeric field`. A `tags` key naming a field column is rejected too. |
| **No test file** | `test_nori_regression.py`: **190 tests**, mock-based in the style of `test_signal_filter.py` (fake `influxdb3_local`, `LineBuilder` and `requests`), no engine and no network. Covers the allowlist and pinning regressions, prediction validation, retry classification, the multi-series guard, window resolution, schema typing, the read ceiling, partial-batch recovery, and which text is allowed to reach a caller. One test asserts the docstring metadata matches the accepted parameter set, and one loads the shipped TOML template, so both classes of drift fail the suite. |

## 5. Other findings

| Item | Resolution |
|---|---|
| `:370` identical payload every tick | `skip_existing`, defaulting to true: the output measurement is read first (filtered by source, target **and** the source series tags, so a sibling trigger cannot interfere) and already-predicted timestamps are dropped. Verified on a real `every:1m` trigger — first tick wrote 4 points, second tick made **zero** gateway calls. Kept as a flag, as you suggested, so newer training data can refresh earlier predictions. |
| `:103` field name with a space | Documented as a limitation of `--trigger-arguments`, with the two forms that do work: a TOML array and a JSON list in the HTTP body. |
| `:397` `str(e)` in the response | Split by intent. `ConfigError` and `GatewayError` messages are written for callers and name no internals — those are returned verbatim, because "internal error" for a bad column name would be hostile. Everything else returns a stable `"internal error; see the plugin logs for this task_id"` with the detail logged. A test asserts a storage failure does not leak `target_database` into the response. |
| `:337` write retry catches everything | This one cannot be fixed the way it is phrased, and I checked before saying so: the engine raises a bare `builtins.Exception` for **every** write failure, transport and schema alike, with only the message to distinguish them (verified on 3.10.3 — a field-type conflict and a connection fault are indistinguishable by type). Classifying on message text would be worse than the wait. So writes go through the repo's shared `write_data` and the knob is the wait: `max_retries=1` writes once and fails fast. The rationale is now in the `_write_predictions` docstring so it does not get re-litigated. I measured your ~5 s: 3.7 s for a permanent conflict at the default. |

## Found while hardening this round

Not from the review, but the same class of thing, so worth listing:

- **A source tag column named `model`, `source` or `target` was quietly corrosive.** Output points carry those three as provenance tags, so such a column would overwrite the provenance *and* make the `skip_existing` lookup contradict itself (`"source" = <tag value> AND "source" = <measurement>`), meaning the run would never skip and would re-pay for the same rows forever. Now rejected up front.
- **A source tag column named `y` made the read query unplannable.** The target was aliased `AS y`, so the alias collided with the column and every run failed with an opaque planner error. The alias is gone; columns are read under their own names.
- **A gateway failure in a later batch discarded the batches already paid for**, and the next run bought them again. A partial run now writes what it has and logs the shortfall.
- A negative `Retry-After` made `time.sleep` raise and abandoned attempts that were still healthy; the delay is clamped at both ends.
- A non-string, non-bytes request body raised `TypeError` straight out of `process_request`, bypassing the sanitised error contract.
- The TOML "supplies all parameters" claim was not enforced — inline arguments alongside `config_file_path` were silently dropped. Now rejected.
- The TOML loader's error echoed the resolved host path to HTTP callers. `PublicError` now carries a `detail` that only reaches the log; the same mechanism keeps the gateway's echoed body out of responses.

## Also changed

- Rebased onto current `main` (the branch was 53 files behind, predating the `validate-registry-plugins` CI and the `influxdata-plugin-utils` convention).
- Dropped `pandas` and `numpy`. Query rows deliver `time` as integer nanoseconds, so plain Python does the whole job; dependencies are now `influxdata-plugin-utils>=0.2.0` and `requests`.
- `output_measurement` equal to `measurement` is rejected (it would write predictions into the column read as ground truth).
- `min_history > max_train_rows` is rejected, since no run could ever qualify.

## Verification

- `pytest influxdata/nori_regression/` — 190 passed.
- `scripts/validate_readme.py --plugins nori_regression` — 0 errors, **0 warnings**.
- `scripts/validate_influxdb3_syntax.py` — 0 errors, 0 warnings.
- `influxdb3-plugin validate influxdata/nori_regression` (SDK v0.6.0, the CI gate) — `{"status":"ok"}`.
- Live on `influxdb:3-core` 3.10.3 with a real key, four series against a known relationship
  `pressure = 1000 + 0.5·temp − 0.3·humidity + noise(sd 0.3)`:

  | series | rows imputed | MAE | exercising |
  |---|---|---|---|
  | A | 20 | 0.105 | plain HTTP run |
  | B | 4 | 0.059 | TOML-driven scheduled trigger, 2 batches |
  | C | 20 | 0.067 | truncated read (`max_read_rows=90`) |
  | D | 15 | 0.035 | body filling what the trigger left open, 3 batches |

  Also fired live: the multi-series guard (two series with disjoint timestamps — the case the old
  guard missed), the body-override and body-pinning rejections, the `gateway_url` and legacy-parameter
  rejections, the retired-slug rejection, the TOML-plus-inline-argument rejection, the read-ceiling
  warning, and a 404 whose body stayed in the log while the caller got the status.

- **Idempotency, soaked.** A real `every:1m` trigger ran ~1,260 times over 21 hours and made
  **12 gateway calls in total** — all on the first tick. Without `skip_existing` that window would
  have re-sent its rows on every one of those ticks.

- **With authentication on** (not `--without-auth`): the engine endpoint returns 401 without a token,
  and every rejection above still fires for a caller holding a valid token. I also isolated the
  `@`-token guard — a copy of the plugin with only that guard removed, same trigger, same token,
  returned the operator's `NORI_API_KEY` and the contents of a planted host file; the real plugin
  refuses both. Worth noting that Core only issues **admin** tokens (resource-scoped tokens are
  Enterprise), so on Core the caller already owns the database — but an admin token still does not
  entitle them to the host's environment or filesystem, which is what made this an escalation.

- **`target_database`** pointed at a second database wrote all 20 predictions there, with the source
  series tags intact, and nothing into the trigger's own database.

- **The retry path, driven for real** against a mock gateway on `127.0.0.1` inside the container
  (which is what the loopback carve-out in `NORI_GATEWAY_URL` exists for):

  | forced condition | observed |
  |---|---|
  | 503, then 429 with `Retry-After: 2`, then 200 | 3 attempts in 3s, 20 written — `Retry-After` honoured |
  | permanent 404 | **1 attempt**, no retry |
  | 503 on every attempt | 3 attempts, then failed |
  | 200 with an HTML body | `non-JSON body (HTTP 200, content-type 'text/html')` |
  | every prediction null | failed, not "success with 0 written" |
  | batch 1 OK, batch 2 404 | `partial`, 8 written, 12 remaining, and the 8 persisted |

  In every failing case the gateway's own text was in the log and not in the response.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed. `_resolve_window` now applies each bound on its own: `start_time` alone reads up to now, `end_time` alone reads one `window` before it, and only when neither is given does the window hang off now. An empty or inverted range is rejected rather than silently widened.

The `INTERVAL` string interpolation is gone entirely too — both bounds are computed in Python and passed as bound query parameters.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Good catch, fixed. An empty or whitespace-only `X-Nori-Api-Key` now falls through to the `NORI_API_KEY` environment variable instead of winning and sending `Api-Key ` with no key. The same path also handles a non-string value (a repeated header delivered as a list) rather than raising `AttributeError` on `.strip()`.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed, and wider than asked — but first: **this had a second half that the key-name allowlist alone does not close.**

The allowlist gates key *names*, while the *values* still flowed into `load_plugin_config`, and dynaconf evaluates converter tokens on any string starting with `@`:

```bash
curl -X POST .../api/v3/engine/nori_regress -d '{"measurement":"@format {env[NORI_API_KEY]}", ...}'
# {"status":"failed","message":"target field 'p' not found in 'SECRET-KEY-abc123'. ..."}
```

That returns the operator's `NORI_API_KEY`. `@read_file /path` returns arbitrary host file contents the same way, which also reinstates the arbitrary file read that keeping `config_file_path` out of the body was meant to prevent. I verified it live with auth enabled and a real token, and isolated it: a copy of the plugin with only the `@`-guard removed leaks; the shipped one refuses.

**It is not specific to this plugin** — `gapfill` on `main` passes its request body straight into the same helper. Details and a reproduction through the helper itself are in [the summary comment](https://github.com/influxdata/influxdb3_plugins/pull/115#issuecomment-5144187159).

What changed here:
- `gateway_url` is no longer a parameter at all (see the thread on `:126`).
- Explicit `BODY_OVERRIDABLE_KEYS`: `measurement`, `field`, `feature_fields`, `tags`, `window`, `start_time`, `end_time`, `dry_run`. Everything else is operator-only and is **rejected by name**, not dropped silently.
- **A trigger argument pins its value.** The allowlist alone was still not enough: `output_measurement` defaults to `<measurement>_regressed` and `measurement` is body-settable, so a body could still move the write target. The body may now only fill in what the trigger left open.
- Values beginning with `@` are refused.

Regression tests: one parametrized case per operator key, per key-name trick (case, whitespace, unicode look-alikes, dunder), and per injection token.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Done, and taken further than the suggestion. `gateway_url` is not a parameter in either place — it is a module constant plus a `NORI_GATEWAY_URL` environment override for operators running a private gateway, which is the same trust level as `NORI_API_KEY` and is not logged with the trigger arguments.

The override must be `https://` (plain `http://` only for a loopback host, so a local mock gateway stays testable) and is parsed with `urlsplit`, so `http://localhost@evil.example/` resolves to `evil.example` and is rejected. A trigger argument named `gateway_url` is now **rejected with a message naming the replacement** rather than silently ignored — as are the parameters from the earlier forecasting revision (`mode`, `horizon`, `step`, `lags`, `rolling`, `tz`).

`target_database` follows the rule you asked for: trigger-argument or TOML only, never the body.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Both parts fixed.

**Destination:** the endpoint is a module constant now, so no input chooses where the key goes. HTTPS is required outside loopback.

**Retry:** the call retries 408/409/425/429 and every 5xx plus connection errors, honours `Retry-After`, and backs off exponentially with jitter capped at 30s. A read timeout is deliberately *not* retried — it has already spent the whole budget, and per Synthefy's own client a timeout usually means the key's group was never granted the slug. `request_timeout` is configurable.

The default is now `300s`, and that matters more than it looks: I measured the cold start at **125.5s on `synthefy/nori-30m`**, above the old fixed 120s, so the previous timeout would have failed every cold start on the model the plugin defaulted to.

Driven for real against a mock gateway on loopback: 503 → 429 (`Retry-After: 2`) → 200 took 3 attempts in 3s; a permanent 404 took exactly 1 attempt.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed. Tag parsing is now `influxdata_plugin_utils.parsing.parse_key_value(pair_sep=" ", kv_sep=":")`, which raises on a token with no separator:

```
tags=site=A  ->  invalid `tags`: Invalid pair 'site=A' (missing ':').
                 Format: 'key:val key2:val2' (space-separated pairs).
```

Verified live.

influxdata/nori_regression/nori_regression.py
ajanbekzat: You were right, and it turns out to be a real cold-start behaviour rather than a hypothetical: a proxy in front of a scaled-to-zero model can answer **200 with an HTML body**.

Body decoding moved into its own `_decode_body` with its own `try`/`except`, so the two failures can no longer be confused. A gateway that returns non-JSON now raises `GatewayError("gateway returned a non-JSON body (HTTP 200, content-type 'text/html')")`, which names the gateway instead of blaming the caller's body. Forced live against a mock that returns exactly that.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed at the gateway boundary, with one nuance worth flagging because it changes your prescription slightly.

A `null` element is **legitimate gateway output**: the model emits JSON `null` for a row whose prediction is not finite (Starlette runs `allow_nan=False`, so null is how non-finite is expressed). Requiring a finite number for *every* element would therefore fail a whole batch over one unpredictable row.

So: a `null` is a per-row outcome — skipped, warned and counted. Anything else non-numeric (a numeric string, a `NaN` literal, a bool) fails the batch, since that breaks the documented contract. And an **all-null** batch fails, which closes the "success with `written: 0`" hole you identified.

If you would rather any null fail the batch, say so and I'll flip it — it's a one-line change.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Both fixed on that line. An empty or whitespace-only header value now falls through to the environment variable instead of winning it, and the value is type-checked before `.strip()` (a repeated header delivered as a list is handled, and a non-string key can't raise). The `Authorization` header is still never read.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed. The message now reads "the space-separated numeric feature columns (X) used to predict `field`", matching the docstring and the README.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed, and made drift-proof. Both `scheduled_args_config` and `http_args_config` now declare all **20** parameters, and each description marks whether the parameter is trigger-argument-only.

There is also a test asserting the declared set is exactly the set of keys the code accepts, so this cannot silently drift again.

influxdata/nori_regression/README.md
ajanbekzat: You were right that it was false. The claim is now true rather than removed: the guard counts **distinct tag combinations** in the window instead of duplicate timestamps, so two series with disjoint timestamps are caught.

```
the window holds 2 series (e.g. {site=A}, {site=B}), but a run must resolve to one: ...
```

Verified live against exactly that shape. I also took your alternative in part — output points now carry **every tag of the source series**, not only the tags the caller filtered on, so a prediction is traceable even when no `tags` filter was given.

influxdata/library/plugin_library.json
ajanbekzat: Respectfully pushing back on this one, with evidence — happy to be overruled.

`REQUIRED_PLUGIN_METADATA.md:55` does say the file is generated and should not be hand-edited, but current practice on `main` is the opposite: every recent plugin PR edits it in the plugin's own commit.

```
$ git show --stat 9f499a7   # gapfill (#128)
 influxdata/library/plugin_library.json | 11 +
$ git show --stat fd06cd5   # resampler (#125)
 influxdata/library/plugin_library.json | 10 +
$ git show --stat 3bd4bbc   # signal_filter (#121)
 influxdata/library/plugin_library.json | 13 +-
```

So I kept the entry and matched that convention instead: fixed the stale date (now `2026-07-30`), mirrored `required_libraries` from `manifest.toml` verbatim, and dropped the retired slug from the description.

If you'd rather the doc win, say so and I'll remove the entry — but then `REQUIRED_PLUGIN_METADATA.md` and the last three merged plugin PRs disagree, which seems worth settling separately.

influxdata/nori_regression/README.md
ajanbekzat: Fair hit — the description reported errors and omitted the warnings.

It is now **0 errors and 0 warnings**. All five missing sections were added (`### Optional parameters`, `### TOML configuration`, `### Data requirements`, `### Software requirements`, `### Schema requirements`), along with an `influxdb3 query` example and an `influxdb3 write` example.

While fixing the examples I found that the one I'd written could never have produced the output it showed — it seeded 3 labeled rows against a `min_history` of 50. That's corrected too.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed as you prescribed: the guard counts distinct tag combinations in the window and fails above one, before any gateway call. Two series with disjoint timestamps no longer slip through.

I also took the second half of your alternative: output points carry the source series' own tags, so a prediction can be traced back to its series even when the caller passed no `tags` filter. Full multi-series-in-one-run remains a possible enhancement and is listed under Limitations.

Verified live with two series at disjoint timestamps — the case the old guard missed.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed, and you were right that caps applied after materialisation are not a bound. I measured your exact scenario before adding one: a 30-day window at 1 Hz peaked at **~1.8 GB** inside the plugin process to send 10,000 rows.

Two layers now:
- **The read** is bounded by a real `ORDER BY time DESC LIMIT` (`max_read_rows`, default 50000 ≈ 35 MB), and a truncated read logs a warning rather than passing silently.
- **The payload** is bounded by `max_train_rows` (1000), `max_predict_rows` (5000) and `predict_batch_size` (1000), keeping the most recent rows and logging what was dropped.

One caveat worth recording: **batching is not free.** Each batch re-sends `X_train`, and billing is per request on `X_train` rows × columns, so splitting `X_test` into N batches bills N times. The log says so and the README has a "Cost and metering" section.

Your two-query suggestion (`IS NOT NULL LIMIT max_train_rows` / `IS NULL LIMIT max_predict_rows`) is strictly more efficient than one capped read, and I'd take it as a follow-up — I left it out here only because it interacts with `skip_existing` in a way that needs care to avoid starving older unpredicted rows.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed. Each bound now applies on its own — `start_time` alone reads up to now, `end_time` alone reads one `window` before it — and an incomplete or inverted range is rejected rather than silently replaced by the default window.

The `INTERVAL` interpolation is gone as well: both bounds are computed in Python and bound as query parameters.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed. `_resolve_schema` now reads `column_name, data_type` from `information_schema.columns` and rejects a non-numeric target or feature up front:

```
feature field 'site' is Dictionary(Int32, Utf8), not a numeric field.
Nori regresses on numeric columns only.
```

A `tags` key that actually names a field column is rejected too. Numeric is the same set `influxdata_plugin_utils.introspection` uses: `Int64`, `UInt64`, `Int32`, `Float64`, `Float32`.

The type read also caught a related problem: a source tag column named `model`, `source` or `target` would collide with the provenance tags on output points — overwriting them *and* making the `skip_existing` lookup self-contradictory, so the run would never skip and would re-pay forever. That configuration is now refused.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed — validation moved to the gateway boundary as you suggested (see the thread on `:241` for the one nuance about legitimate nulls).

`_run` no longer treats a truthy list as proof of work. It also now distinguishes a **partial** run: if a gateway fault stops the run part-way, the batches that already returned are still written — they were already billed — and the result reports the shortfall rather than `success`:

```json
{"status": "partial", "written": 8, "remaining": 12}
```

Verified live by forcing a 404 on the second of three batches.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Taken, and it changed more than I expected. `skip_existing` (default true) reads `output_measurement` first and drops timestamps that already hold a prediction, filtered by source, target **and** the source series tags so a sibling trigger writing another series can't interfere. Kept as a flag, as you suggested, so later training data can refresh earlier predictions.

Incidental soak evidence: a real `every:1m` trigger ran ~1,260 times over 21 hours and made **12 gateway calls in total**, all on the first tick.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Documented. The README notes that every string form splits on whitespace, so a column name containing a space is unreachable from `--trigger-arguments`, and gives the two forms that do work: a JSON list in the HTTP body (as you noted), and — new in this round — a TOML array via `config_file_path`:

```toml
feature_fields = ["air temp", "humidity"]
```

That fell out of implementing `config_file_path`, so the scheduled trigger now has a way to express it too.

influxdata/nori_regression/nori_regression.py
ajanbekzat: Fixed here, structurally rather than case by case. There is now a `PublicError` base with a `detail` field: the message is written for the caller, and anything that must not travel goes in `detail`, which only reaches the log.

- `ConfigError` / `GatewayError` messages are returned verbatim — they name the caller's own input, and returning "internal error" for a bad column name would be hostile.
- The gateway's echoed body, the resolved TOML path, and any other exception now go to the log, with the caller getting a stable message plus the `task_id` to correlate on.

```
caller:  Nori gateway returned HTTP 404 for model 'synthefy/nori-30m'; see the plugin logs for this task_id
log:     GatewayError: ... [detail: please check the model you provided]
```

A test asserts a storage failure doesn't leak `target_database` into the response. Agreed that the broader pattern is repository-wide.

influxdata/nori_regression/nori_regression.py
ajanbekzat: This one can't be fixed the way it's phrased, and I checked before saying so: the engine raises a bare `builtins.Exception` for **every** write failure — a field-type conflict and a connection fault are indistinguishable by type, only by message text (verified on 3.10.3). Classifying on message text would be worse than the wait.

So writes go through the repo's shared `influxdata_plugin_utils.write.write_data` and the knob is the wait instead: `max_retries=1` writes once and fails fast. I measured your ~5s at 3.7s for a permanent conflict at the default. The rationale is now in the `_write_predictions` docstring so it doesn't get re-litigated.

If the engine ever grows typed write exceptions, narrowing belongs in the shared helper so every plugin gets it.

caterryan (approved):

caterryan:
I created a separate issue for the security issue in influxdata-plugin-utils
https://github.com/influxdata/influxdb3_plugins/issues/134

caterryan:
The security issue in influxdata-plugin-utils https://github.com/influxdata/influxdb3_plugins/issues/134 was fixed in https://github.com/influxdata/influxdb3_plugins/pull/135

Not required for this PR, but bump the version to 0.3.0 to get the latest.

Aliaksei-Kharlap (approved):

PR #128, opened 2026-07-21
feat: add gapfill plugin

Aliaksei-Kharlap:


caterryan (commented):
Manually validated this plugin end-to-end on a live InfluxDB 3 Enterprise instance.

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+    numeric_fields: list[str] = []
+    line_types: dict[str, str] = {}
+    for row in rows:
+        name, data_type = row["column_name"], row.get("data_type", "")
+        if data_type == TAG_DATA_TYPE:
+            tags.append(name)
+            continue
+        if name == "time":
+            continue
+        if marker is not None and name == marker:
```
caterryan, line 230: **[P1] In-place `mark_filled` works exactly once, then every later run fails.**

The first in-place run writes the marker field into the source measurement; the next run's `resolve_schema` sees that column and raises the collision error. Reproduced live: run 1 filled a gap and wrote `filled=true`; run 2 with identical args (and a new gap present) returned `source_measurement 'f01_rerun' already has a field named 'filled'` — the new gap is never repaired. A scheduled in-place trigger with `mark_filled=true` permanently errors after its first fill.

Fix: in in-place mode, accept an existing Boolean column named `filled_field_name` (it is plugin-owned) and exclude it from fill/carry inputs instead of raising.

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+                cls = CubicSpline if method == "cubic" else PchipInterpolator
+                interpolator = cls(x, vals)
+
+        for start_ns, end_ns in gaps:
+            # Nodes sit on the epoch-anchored grid strictly inside the gap, so
+            # overlapping runs overwrite the same points idempotently. Chain
+            # mode clips at node_start_ns so fills stay inside the copied
+            # window.
+            first: int = (start_ns // interval_ns + 1) * interval_ns
+            if not cfg["in_place"] and first < node_start_ns:
```
caterryan: **[P1] HTTP in-place backfill writes outside the requested range.**

The lower clip to `node_start_ns` applies only in chain mode, so an in-place backfill fills a gap all the way back to its start even when that precedes `backfill_start`. Reproduced live: gap 09–20s, `backfill_start=...:00:15Z` → fill points written at 10–14s, before the requested range. Same applies at HTTP batch boundaries.

Fix: clip HTTP-mode fills to `[backfill_start, backfill_end)` regardless of output mode (the scheduled trigger's unclipped in-place behavior is fine — an explicit HTTP range should be honored).

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+                first = node_start_ns
```
caterryan: **[P1] `first = node_start_ns` shifts fills off the epoch grid, breaking idempotency.**

When `node_start_ns` isn't a multiple of `interval_ns` (any unaligned `backfill_start`, batch boundary, or scheduled window edge), this assignment makes every node in the gap off-grid. Reproduced live: chain mode, `interval=1s`, `backfill_start=...:00:15.500Z` → fills at 15.5/16.5/17.5/18.5/19.5s; re-running with an aligned start then added on-grid 10–19s, leaving **both phases permanently interleaved** in the target measurement. This directly contradicts the README's "epoch-anchored grid … overlapping runs are idempotent".

Fix: ceil to the grid instead of assigning the raw boundary, e.g. `first = max(first, -(-node_start_ns // interval_ns) * interval_ns)` (same idiom as resampler's `compute_grid`).

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+                j = bisect.bisect_right(times, ts) - 1
+                if j >= 0:
+                    fill_points[ts][field] = values_l[j]
+
+    lines: list = []
+    for ts in sorted(fill_points):
+        values: dict = fill_points[ts]
+        if cfg["mark_filled"]:
+            values[cfg["filled_field_name"]] = True
+        if field_types is None:
```
caterryan: **[P1] Chain mode degrades UInt64 columns to Int64.**

The chain path uses `build_line`, which infers every Python int as signed Int64. Reproduced live: source `count` UInt64 (`100u`), chain mode with `method=previous` → target column created as `Int64` (verified via `information_schema.columns`). Values above INT64_MAX would fail the write outright. The in-place path already solves this with `build_line_typed`.

Fix: thread the resolved schema's line types through the chain fill and copy paths as well, overriding method-filled fields to float for blending methods.

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+                            if ftype == "uint"
+                            else (INT64_MIN, INT64_MAX)
+                        )
+                        if not (lo <= value <= hi_v):
+                            # Outside the column's integer range: skip the
+                            # node, not the whole run.
+                            out_of_range_fields.add(field)
+                            continue
+                    fill_points.setdefault(ts, {})[field] = value
+                stats["points"] += n_writable
```
caterryan: **[P2] `fills_written` and the gap report count nodes that were never written.**

`stats["points"] += n_writable` runs even when the integer-range check `continue`s every node, and the report row uses `n_writable` too. Reproduced live: Int64 column, `method=constant, fill_value=9.3e18` (> INT64_MAX) → response said `gaps_filled=1, fills_written=10` and the report row said `status=filled, points=10`, but **zero rows were written** (the warn log was the only honest signal).

Fix: count successful assignments, and report `status=skipped` / `points=0` when all nodes were dropped.

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+        "WHERE table_name = $table",
+        {"table": measurement},
+    )
+    tags: list[str] = []
+    all_fields: list[str] = []
+    numeric_fields: list[str] = []
+    line_types: dict[str, str] = {}
+    for row in rows:
+        name, data_type = row["column_name"], row.get("data_type", "")
+        if data_type == TAG_DATA_TYPE:
```
caterryan: **[P2] Marker name colliding with a tag passes validation, then fails opaquely at write time.**

Tag columns `continue` here before the marker-collision check below, so `filled_field_name=<tag name>` validates cleanly. Reproduced live with `filled_field_name=room`: the run reached the write and died with `invalid column type for column 'room', expected iox::column_type::tag, got iox::column_type::field::boolean` instead of the clear collision message.

Fix: include tag names in the marker-collision check.

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+    elif fill_value_raw is not None:
+        raise Exception(f"[{task_id}] fill_value is only valid with method 'constant'.")
+
+    mark_filled: bool = parse_bool_arg(cfg.get("mark_filled"))
+    filled_field_name: str = str(cfg.get("filled_field_name") or "filled").strip()
+    if not filled_field_name or filled_field_name == "time":
+        raise Exception(
+            f"[{task_id}] filled_field_name must be a non-empty field name."
+        )
+    report: str | None = cfg.get("report_measurement")
```
caterryan: **[P2] Whitespace `report_measurement` aborts the entire run and drops the computed fills.**

`target_measurement` and `target_database` are normalized with `.strip() or None`, but `report_measurement` isn't. Reproduced live with `report_measurement=" "`: LineBuilder raised `Measurement name cannot contain spaces` while building report lines, the whole run errored, and the 10 already-computed fill points were lost — a stray space disables filling entirely, not just the report.

Fix: apply the same `str(...).strip() or None` normalization here.

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+
+    if builders:
+        write_data(
+            influxdb3_local,
+            builders,
+            retries=cfg["max_retries"] - 1,
+            no_sync=True,
+            database=cfg["target_database"],
+        )
+    if report_lines:
```
caterryan, line 816: **[Docs] Gap report follows `target_database` — undocumented.**

With chain mode + `target_database`, report rows are written to the target DB. Verified live: the report measurement appeared only in the target database, not the trigger's. Defensible design, but the README's monitoring example (notifier alerting on the report) would silently watch the wrong database. Either route reports to the trigger DB or add a README sentence stating where they land.

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
+"""
+{
+    "plugin_type": ["scheduled", "http"],
```
caterryan, line 3: **[P2] Docstring metadata missing `http_args_config`.**

`plugin_type` declares `"http"`, but only `scheduled_args_config` is present, so Explorer cannot configure the HTTP trigger (`backfill_start`, `backfill_end`, `batch_size`, plus the shared params). Per REQUIRED_PLUGIN_METADATA.md a config array is omitted only when the trigger type is unused; downsampler and prophet_forecasting both carry `http_args_config` for this case. Add it (shared params minus `window`/`offset`, plus the three HTTP-only params).

influxdata/gapfill/gapfill.py
```diff
@@ -0,0 +1,985 @@
...
+from influxdata_plugin_utils.config import Validator, load_plugin_config
+from influxdata_plugin_utils.introspection import query_window
+from influxdata_plugin_utils.parsing import (
+    parse_bool,
+    parse_delimited_list,
+    parse_timedelta,
+    parse_timestamp_ns,
+)
+from influxdata_plugin_utils.write import (
+    _infer_type,
```
caterryan: **[P2] `_infer_type` is a private utils API.**

It is importable but not exported — `influxdata_plugin_utils.write.__all__` is `['add_field_with_type', 'build_line', 'build_line_typed', 'write_data', 'BatchLines']` (verified against 0.2.0). With the open-ended pin `>=0.2.0`, plugin environments resolve the latest utils, so a future rename breaks deployed triggers at import time. Export it from the utils package, or inline the small type-inference fallback here.

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
Aliaksei-Kharlap: fixed

influxdata/gapfill/gapfill.py
caterryan: leave for now. we'll fix in a separate pr

caterryan (approved):

PR #132, opened 2026-07-28
resolve nsu= URIs in browse_root, add browse_cache_ttl

### LINKED ISSUE #130: [influxdb3_plugins/influxdata/opcua] nsu= URI addressing silently fails in browse_root
## 1. `nsu=` URI addressing silently fails in `browse_root`

**Severity: high — silent data loss, and it defeats the documented
mechanism for restart stability.**

### Observed

A browse-mode config using URI addressing for the root:

```toml
[opcua.browse]
browse_root = "nsu=urn:influxdata:reference:isa95;s=Site_Liverpool"
browse_depth = 5
path_tags = ["area", "line", "unit"]
```

produces, every cycle:

```
INFO  Connected to OPC UA server successfully
INFO  Browsing OPC UA address space from:
      nsu=urn:influxdata:reference:isa95;s=Site_Liverpool,
      browse_depth=5, path_tags=['area', 'line', 'unit']
INFO  Browse complete: 0 groups, 0 variables discovered
WARN  Browse discovered no variables, reconnecting...
INFO  Reconnecting to OPC UA server...
INFO  Connected to OPC UA server successfully
WARN  Browse still empty after reconnect
INFO  finished execution in 46ms
```

The connection succeeds. The reconnect succeeds. The browse returns
nothing, indefinitely, with no error raised and no indication of the
actual cause. We spent a considerable amount of time on connectivity and
depth hypotheses before reading the source.

Changing only the root to index form fixes it completely:

```toml
browse_root = "ns=3;s=Site_Liverpool"
```

```
INFO  Browse complete: 36 groups, 142 variables discovered
INFO  Data write complete (browse mode): 112 fields written, 0 errors
```

### Cause

`_resolve_namespace_uris` iterates over `("nodes", "tag_nodes")` only. Its
own docstring says "Scans all node IDs in 'nodes' and 'tag_nodes' config
sections." `browse.browse_root` is never processed.

The unresolved string is then passed to `client.get_node(root_node_id)`,
which cannot parse `nsu=...;s=...` as a NodeId. The resulting node object
has no children, the walk terminates immediately, and no exception
propagates.

### Why this matters beyond the workaround

`nsu=` exists precisely because namespace *indices* are assigned at
runtime and can change between server restarts, upgrades, or
configuration changes. The plugin's own documentation recommends URI
addressing for this reason.

The consequence is that explicit mode can be made restart-stable and
browse mode cannot. Browse mode is also the mode that matters at customer
scale — nobody hand-maintains a node list for a few thousand tags. So the
feature most needed at scale is the one pinned to an unstable identifier,
and when that identifier shifts the failure is silent rather than loud.

### Suggested fixes, in order of value

1. Resolve `nsu=` in `browse_root` — add `browse` to the sections
   `_resolve_namespace_uris` walks.
2. Failing that, **fail loudly**: if `browse_root` starts with `nsu=`,
   raise a config error at validation time rather than browsing into
   nothing.
3. Independently, treat "browse returned 0 variables" as an error rather
   than a warning-and-retry. A misconfigured root and a genuinely empty
   subtree are both worth surfacing, and the current message points the
   operator at connectivity, which is misleading.

### LINKED ISSUE #131: [influxdb3_plugins/influxdata/opcua] No browse caching - the tree is re-walked every cycle
## No browse caching — the tree is re-walked every cycle

**Severity: medium now, high at customer scale.**

### Observed

Per-cycle execution time, 10s interval:

| Mode | Nodes | Per-cycle |
|---|---|---|
| Explicit | 15 | ~20 ms |
| Browse (flat model, depth 2) | 160 | ~120 ms |
| Browse (ISA-95 model, depth 5) | 142 | ~120–160 ms |

Roughly 6–8× the cost of explicit mode. At this scale it is irrelevant —
under 2% of the interval.

### The concern

Discovery is repeated in full on every trigger run. There is no
discovery cache, no TTL, and no way to browse every *n*th cycle. It is
browse mode or explicit mode, per trigger, with no middle ground.

Three factors differ between this testbed and a customer deployment, and
they do not scale together:

1. **Node count.** 5,000 tags versus 142 is roughly 35×.
2. **Tree shape.** Browse cost tracks the number of Browse service calls,
   which depends on branching factor, not leaf count.
3. **Network latency.** This testbed runs over loopback (~0.05 ms). A
   plant OPC UA server across a WAN link is commonly 20–50 ms per round
   trip, and browse is inherently chatty.

The third usually dominates. A customer with a deep tree over a slow link
could plausibly spend more time browsing than collecting.

### Comparison

Most commercial collectors discover once,
cache the resolved node set, then poll or subscribe against it,
re-discovering periodically. That is the established pattern and it is
what makes browse mode viable at scale.

### Suggested fix

A `browse_interval` or `browse_cache_ttl` parameter: discover on first
run and every *n* seconds thereafter, reading the cached node set in
between. This would let a customer run discovery every 5 minutes and
collection every 10 seconds — which is what they actually want, and
currently impossible without maintaining two configs.

### Confidence

High on the measurement, high on the absence of caching (confirmed in
source). The extrapolation to customer scale is reasoned, not measured —
we did not have a large or high-latency address space to test against.

Aliaksei-Kharlap:
- resolve nsu=<uri> in browse_root (not only nodes/tag_nodes) at connection time and re-resolve on every reconnect, so namespace index shifts after a server restart are handled, resolution no longer mutates the cached config
- add browse_cache_ttl to cache the discovered browse structure, tie the config cache TTL to it
- add config_cache_ttl 

Closes #131 
Closes #130

### DIFF
--- a/influxdata/opcua/README.md
+++ b/influxdata/opcua/README.md
@@ -90,6 +90,8 @@ The plugin resolves `nsu=` URIs to numeric `ns=` indexes after connecting to the
 | `name_separator`   | string | none      | `[opcua.browse]` | Separator for splitting Variable browse names into segments for tag extraction. Required when `name_tags` is set. See [Name tags](#name-tags).                                                                                                                                                                                                               |
 | `name_tags`        | list   | none      | `[opcua.browse]` | Tag names extracted from leading segments of Variable browse names split by `name_separator`. Remaining segments form the field name (joined with `_`). See [Name tags](#name-tags).                                                                                                                                                                         |
 
+> **Note:** integer parameters (`browse_depth`, `browse_cache_ttl`) also accept numeric strings, so `browse_depth = "2"` is valid in TOML. `browse_cache_ttl` is capped at 30 days.
+
 ### Tag parameters
 
 | Parameter      | Type   | Default | TOML Section          | Description                                                                                                                                                                                                       |
@@ -143,7 +145,8 @@ quality_filter = ["good", "uncertain"]
 | Parameter              | Type   | Default | TOML Section | Description                                                                                                                                                                                                              |
 |------------------------|--------|---------|--------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `config_file_path`     | string | none    | CLI only     | Path to TOML config file (absolute or relative to `PLUGIN_DIR`)                                                                                                                                                          |
-| `disable_config_cache` | bool   | false   | `[opcua]`    | Reload configuration on every call instead of caching for 1 hour. Also disables caching of the discovered browse structure, so in browse mode the address space is re-walked on every call. Useful during development.   |
+| `config_cache_ttl`     | int    | 3600    | `[opcua]`    | Seconds to cache the parsed configuration (max 2592000). Controls how quickly credential, endpoint, table, or filter changes take effect. Independent of `browse_cache_ttl`.                                             |
+| `disable_config_cache` | bool   | false   | `[opcua]`    | Reload configuration on every call instead of caching for `config_cache_ttl` seconds. Also disables caching of the discovered browse structure, so in browse mode the address space is re-walked on every call. Useful during development. |
 | `allow_insecure_auth`  | bool   | false   | `[opcua]`    | Permit sending `username`/`password` when `security_policy` is not set (credentials sent in cleartext over an unencrypted connection). Only enable on a trusted network.                                                 |
 | `enable_full_logging`  | bool   | false   | `[opcua]`    | When `true`, full exception messages are written to logs. When `false` (default), only the exception type is logged, to avoid leaking sensitive values (credentials, payloads, paths). Enable temporarily for debugging. |
 
@@ -319,7 +322,7 @@ If a relative path is specified and `PLUGIN_DIR` is not set, the plugin will ret
 
 | TOML section           | Description                                                                                                                                                      |
 |------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `[opcua]`              | Required. `server_url`, `table_name`, `quality_filter`, `disable_config_cache`, `allow_insecure_auth`                                                            |
+| `[opcua]`              | Required. `server_url`, `table_name`, `quality_filter`, `config_cache_ttl`, `disable_config_cache`, `allow_insecure_auth`                                        |
 | `[opcua.default_tags]` | Static tags as `key = "value"` pairs                                                                                                                             |
 | `[opcua.namespaces]`   | Namespace alias mappings as `alias = "uri"` pairs. Validated but not used for alias substitution in TOML — use `nsu=<uri>;...` directly in node IDs              |
 | `[opcua.tag_nodes]`    | Dynamic tags from OPC UA nodes                                                                                                                                   |
@@ -776,12 +779,17 @@ If all variables in a device group are listed in `browse_tags` (no regular field
 
 ### Discovery caching
 
-The parsed configuration is cached for **1 hour**, and the discovered browse structure is cached for `browse_cache_ttl` seconds (default **1 hour**). Discovery runs on the first call and every `browse_cache_ttl` seconds thereafter; cached node IDs are read in between. This means:
-- Set a low `browse_cache_ttl` for a fast-changing address space, or a high one to run discovery infrequently (e.g. every 5 minutes) while collecting on a short trigger interval — the recommended pattern for large or high-latency servers where a full browse is expensive
-- Changes to the TOML configuration file will not take effect until the cache expires (or use `disable_config_cache = true` during development)
-- New devices added to the OPC UA server will not appear until the browse cache expires
-- A `browse_cache_ttl` longer than 1 hour also extends the config cache to the same duration, so that discovery is not forced sooner by config expiry. As a result, configuration changes take up to `browse_cache_ttl` seconds to apply — set `disable_config_cache = true` to reload immediately during development
-- `disable_config_cache = true` also disables browse caching, so the address space is re-walked on every call
+Two independent caches, each with its own TTL:
+
+- **Parsed configuration** — cached for `config_cache_ttl` seconds (default **1 hour**). Controls how quickly a change to credentials, endpoint, table, or filters is picked up.
+- **Discovered browse structure** — cached for `browse_cache_ttl` seconds (default **1 hour**). Discovery runs on the first call and every `browse_cache_ttl` seconds thereafter; cached node IDs are read in between.
+
+This means:
+- Set a low `browse_cache_ttl` for a fast-changing address space, or set it much larger than the trigger interval so discovery runs rarely while collection stays frequent — the recommended pattern for large or high-latency servers where a full browse is expensive
+- New devices added to the OPC UA server appear after at most `browse_cache_ttl` seconds
+- Changes to the config take effect after at most `config_cache_ttl` seconds. Changes to the browse-relevant part (`server_url`, `browse_root`, `browse_depth`, `filter`, `exclude_branches`, …) also trigger an immediate re-browse on the next config reload; other changes keep the cached structure
+- A failed connection drops the cached config, so a credential or endpoint fix is retried on the next call rather than after `config_cache_ttl`
+- `disable_config_cache = true` disables both caches, so config and address space are re-read on every call
 - Any unhandled error clears all caches, forcing a fresh reload on the next call
 
 ### Namespace URI resolution
@@ -887,7 +895,7 @@ When using a security policy, both files are required:
 
 #### Config changes not taking effect
 
-- Configuration is cached for 1 hour — either wait for expiry or set `disable_config_cache = true`
+- Configuration is cached for `config_cache_ttl` seconds (default 1 hour) — either wait for expiry, lower `config_cache_ttl`, or set `disable_config_cache = true`
 - Any plugin error automatically clears the cache
 
 #### "Previous opcua call still running, skipping this tick"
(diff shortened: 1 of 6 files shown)


influxdata/opcua/opcua.py
```diff
@@ -228,6 +236,20 @@ def _exc(e: BaseException) -> str:
     return str(e) if _ENABLE_FULL_LOGGING else type(e).__name__
 
 
+def _split_nsu_node_id(node_id: str) -> tuple[str, str]:
+    """Split 'nsu=<uri>;<identifier>' into (uri, identifier).
+
+    Raises ValueError if the ';<identifier>' part is missing or empty.
+    """
+    parts = node_id[4:].split(";", 1)
+    if len(parts) != 2 or not parts[1].strip():
```
caterryan: Two problems in this function.

1. The URI keeps its percent-encoding. OPC 10000-6 §5.1.12 requires `;` and `%` in the URI to be percent-encoded in the string form, but `get_namespace_index()` does an exact match against the decoded `NamespaceArray`. A URI that contains `%3B` can never resolve. Decode with `urllib.parse.unquote(parts[0])` before you return it.

2. An empty URI passes validation. `nsu=;s=Devices` fails only after the connection opens. Check both parts:

```python
if len(parts) != 2 or not parts[0].strip() or not parts[1].strip():
```

Please add both cases to the tests.

influxdata/opcua/opcua.py
```diff
@@ -618,6 +641,32 @@ def _validate_toml_nodes(nodes: Any, section_name: str = "opcua.nodes"):
...
+            raise ValueError(
+                f"Parameter '{param_name}' must be a positive integer, got: {value}"
+            )
+        try:
+            parsed = int(value)
+        except ValueError:
+            raise ValueError(
+                f"Parameter '{param_name}' must be a positive integer, got: {value}"
+            )
+        if parsed < 1:
```
caterryan, line 694: Add an optional upper bound to this validator and apply it to the TTL parameters. `browse_cache_ttl = 10**30` passes this check. `cache.put()` then raises `cannot convert float seconds to Duration`, the top-level handler deletes all caches, and the next call repeats the same failure. The plugin never writes data again, and the default log shows only `ValueError`. A 30-day ceiling gives the user a clean message instead of a permanent failure loop.

influxdata/opcua/opcua.py
```diff
@@ -789,15 +844,15 @@ def _build_config_from_args(self) -> dict[str, Any]:
 
             browse_depth = self.args.get("browse_depth")
```
caterryan, line 882: CLI mode does not validate the `browse_root` format. The TOML path calls `_split_nsu_node_id` (L684-685); this path only checks for emptiness, so a malformed value such as `nsu=urn:x` (no `;`) opens a full session before it fails. Add after L843:

```python
if browse_root.startswith("nsu="):
    _split_nsu_node_id(browse_root)
```

influxdata/opcua/opcua.py
```diff
@@ -1814,8 +1892,14 @@ async def _async_scheduled_call(
...
             if not str(cached_config.get("disable_config_cache", False)).lower() == "true":
+                # Keep the config cached at least as long as the browse structure,
+                # so a long browse_cache_ttl is not capped by config expiry (a
+                # config reload invalidates the browse structure below).
+                browse_ttl = (cached_config.get("browse") or {}).get(
+                    "browse_cache_ttl", _DEFAULT_BROWSE_CACHE_TTL
+                )
                 influxdb3_local.cache.put(
-                    "opcua_config", cached_config, _CONFIG_CACHE_TTL
+                    "opcua_config", cached_config, max(_CONFIG_CACHE_TTL, browse_ttl)
```
caterryan: This `max()` couples two unrelated lifetimes: a long `browse_cache_ttl` also delays credential, endpoint, table, and filter changes for the same period. The coupling exists only because L1906 deletes the browse structure on every config reload.

Preferred fix: store a fingerprint of the browse section (`json.dumps(browse_config, sort_keys=True)`) next to the browse cache, keep the cache when the fingerprint and `server_url` are unchanged, and remove the `max()` so each TTL stands alone.

If that does not fit this PR, do two smaller things: delete `opcua_config` when `connect()` fails (L1957), so a credential rotation self-heals on the next call instead of after the full TTL; and add a comment that marks this coupling as a work-around, not a feature.

influxdata/opcua/opcua.py
```diff
@@ -1878,7 +1962,25 @@ async def _async_scheduled_call(
             # Reuse existing connection — update task_id and config for this call
             opcua_client.task_id = task_id
             if not config_from_cache:
-                opcua_client.config = cached_config
+                # Fresh config has unresolved nsu= URIs — re-resolve on the live connection.
+                opcua_client._original_config = cached_config
```
caterryan: Module code writes a private attribute here, and the write is only valid when `_resolve_namespace_uris()` follows it. The two steps are one operation — put them behind one method:

```python
async def apply_config(self, config: dict[str, Any]) -> None:
    """Replace the source config and resolve nsu= URIs against the live session."""
    self._original_config = config
    await self._resolve_namespace_uris()
```

influxdata/opcua/opcua.py
```diff
@@ -1918,12 +2020,21 @@ async def _async_scheduled_call(
...
+                            f"browse_root does not exist on the server (wrong or stale "
+                            f"ns= index), 'filter'/'exclude_branches' excludes every "
+                            f"variable, browse_depth is too shallow to reach Variable "
+                            f"nodes, or the subtree genuinely has no variables."
                         )
                         return
+                browse_cache_ttl: int = browse_config.get(
+                    "browse_cache_ttl", _DEFAULT_BROWSE_CACHE_TTL
+                )
                 influxdb3_local.cache.put(
```
caterryan: Two small items. With `disable_config_cache = true`, L1906 deletes this entry at the start of every call, so this write is wasted — skip the put in that mode. Separately, `_DEFAULT_BROWSE_CACHE_TTL` is applied as a fallback at two read sites (here and L1899) and the two can drift — normalize the default once during validation instead.

influxdata/opcua/README.md
```diff
@@ -773,17 +774,22 @@ Unlike `tag_nodes` which always require "good" quality, `browse_tags` variables
 
 If all variables in a device group are listed in `browse_tags` (no regular fields remain), no data point is written for that group. InfluxDB requires at least one field per point — browse_tags alone are not sufficient.
 
-### Config and browse structure caching
+### Discovery caching
 
-Both the parsed configuration and the browse structure are cached for **1 hour** (controlled by `_CONFIG_CACHE_TTL`). This means:
+The parsed configuration is cached for **1 hour**, and the discovered browse structure is cached for `browse_cache_ttl` seconds (default **1 hour**). Discovery runs on the first call and every `browse_cache_ttl` seconds thereafter; cached node IDs are read in between. This means:
+- Set a low `browse_cache_ttl` for a fast-changing address space, or a high one to run discovery infrequently (e.g. every 5 minutes) while collecting on a short trigger interval — the recommended pattern for large or high-latency servers where a full browse is expensive
```
caterryan: Non-blocking. This sentence is easy to misread: "a high one ... (e.g. every 5 minutes)" reads as a contradiction next to the 3600 s default. The intended comparison is against the trigger interval. Suggest: "set `browse_cache_ttl` much larger than the trigger interval, so that discovery runs rarely while collection stays frequent".

caterryan (changes requested):
The core design is sound. I verified: the deep copy keeps the original `nsu=` config, `uri_to_index` is fully built before any node ID changes, and each reconnect re-resolves against the live namespace array. The items below block merge. Two of them are on lines outside the diff, so they are here in the body.

**1. Error masking — fix this first.** `opcua.py` L467-468 sets `_ENABLE_FULL_LOGGING` from the config key before `_validate_toml_nodes` (L495) and `_validate_toml_browse` (L499) run. When the key is absent, the value becomes `False`, so every later validation error logs only `ValueError`. This defeats the default-True intent that the comment at L229-231 states. The CLI path has the same problem at L784-788: the assignment runs before the rest of argument parsing. Move both assignments to after validation completes. This bug hides the error text for every other problem in this list.

**2. Raw exception text at L1866.** `_read_with_reconnect` logs `{e}` directly. Use `_exc(e)` so the log does not leak details when full logging is off. The PR touches these paths, and this is the last raw-exception log in them.

**3. Add a minimal test file.** This PR adds cache-lifecycle and namespace-resolution logic, and four sibling plugins already have a `test_*.py`. Minimum set: `_split_nsu_node_id` cases (include a `%3B` URI and an empty URI), `_validate_positive_int` bounds, and one assertion with a fake cache that `cache.put("opcua_config", ...)` receives the expected TTL.

**4. Document the TOML coercion change.** `_validate_positive_int` coerces numeric strings, so `browse_depth = "2"` in TOML is now accepted where main rejects it. Add one README line that states this, so the behavior change is deliberate and visible.

influxdata/opcua/opcua.py
Aliaksei-Kharlap: fixed

influxdata/opcua/opcua.py
Aliaksei-Kharlap: fixed

influxdata/opcua/opcua.py
Aliaksei-Kharlap: fixed

influxdata/opcua/opcua.py
Aliaksei-Kharlap: fixed

influxdata/opcua/opcua.py
Aliaksei-Kharlap: fixed

influxdata/opcua/opcua.py
Aliaksei-Kharlap: fixed

influxdata/opcua/README.md
Aliaksei-Kharlap: fixed

caterryan (approved):

PR #139, opened 2026-08-05
[feat]: refactor valuecounter to use utils package

Aliaksei-Kharlap:
Rewrote valuecounter to use influxdata-plugin-utils. The only breaking change is the installation of an additional dependency. New version: "1.0.0"

### DIFF
--- a/influxdata/valuecounter/manifest.toml
+++ b/influxdata/valuecounter/manifest.toml
@@ -2,7 +2,7 @@ manifest_schema_version = "1.2"
 
 [plugin]
 name = "valuecounter"
-version = "0.2.0"
+version = "0.3.0"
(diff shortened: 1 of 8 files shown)


influxdata/valuecounter/manifest.toml
```diff
@@ -2,7 +2,7 @@ manifest_schema_version = "1.2"
 
 [plugin]
 name = "valuecounter"
-version = "0.2.0"
+version = "1.0.0"
```
caterryan:
```suggestion
version = "0.3.0"
```

caterryan (changes requested):

influxdata/valuecounter/manifest.toml
Aliaksei-Kharlap: Done

caterryan (approved):

PR #149, opened 2026-08-20
Adding a new plugin: Earthquake Plugin, ingests earthquake events and writes normalized points for dashboards and alerting

rkinginflux:
## Summary
- Adds a new scheduled plugin: `influxdata/earthquake_sampler/earthquake_sampler.py`
- Supports USGS GeoJSON ingestion and writes normalized earthquake points for dashboards and alerting
- Adds optional canonical `quake` schema output mode (`write_quake_schema=true`) for direct writes into existing `quake` tables
- Adds plugin docs: `influxdata/earthquake_sampler/README.md`
- Registers plugin in `influxdata/library/plugin_library.json`

## Validation
- `python3 -m py_compile influxdata/earthquake_sampler/earthquake_sampler.py`
- `python3 -m json.tool influxdata/library/plugin_library.json`

### DIFF
--- a/influxdata/earthquake_sampler/README.md
+++ b/influxdata/earthquake_sampler/README.md
@@ -1,112 +1,249 @@
 # InfluxDB 3 Earthquake Sampler Plugin
 
-A scheduled InfluxDB 3 Processing Engine plugin that ingests earthquake events and writes normalized points for dashboards and alerting.
+⚡ scheduled 🏷️ earthquake, geoscience, sample-data, monitoring, alerting 🔧 InfluxDB 3 Core, InfluxDB 3 Enterprise
 
-## What this plugin does
+## Description
 
-On each scheduled run, the plugin:
+The Earthquake Sampler Plugin ingests earthquake events from the USGS GeoJSON feeds on a schedule and writes normalized points for dashboards and alerting. It can also read from a custom JSON endpoint or from an existing InfluxDB table, and optionally writes directly into an existing canonical `quake` table using that table's column names (`write_quake_schema=true`). Events are deduplicated with a per-event update-marker cache, so reruns only write new or updated earthquakes.
 
-- reads earthquake data from one of two source modes:
-  - `source_type=http`:
-    - USGS GeoJSON feeds (`feed=all_hour`, `significant_day`, `4.5_week`, etc.), or
-    - a custom JSON endpoint (`source_url=...`) with parser `source_format=usgs_geojson|flat_json`
-  - `source_type=influxdb_table`:
-    - reads rows directly from an existing InfluxDB table (for example `usgs.quake`)
-- normalizes each event into a common schema
-- filters by minimum magnitude (`min_magnitude`)
-- enforces per-run cap (`max_events`)
-- deduplicates events using cached update markers (`skip_unchanged=true`)
-- writes normalized earthquake event rows to the destination measurement (default `earthquakes`)
-- can write directly into an existing canonical `quake` table with its original column names and types (`write_quake_schema=true`)
+- **Zero authentication**: USGS feeds require no API keys or signup
+- **Two source modes**: HTTP (USGS or custom JSON) and `influxdb_table` (transform rows from an existing table)
+- **Per-event deduplication**: update markers are cached per event id; updated events are re-written, unchanged events are skipped
+- **Canonical quake schema**: optional direct writes into an existing `quake` table
 
-## Files
+## Configuration
 
-- `earthquake_sampler.py` - plugin source code
+Plugin parameters may be specified as key-value pairs in the `--trigger-arguments` flag (CLI) or in the `trigger_arguments` field (API) when creating a trigger. All parameters are optional; invalid values abort the run with an error log rather than silently falling back to defaults.
 
-## Key trigger arguments
+### Plugin metadata
 
-- `source_type`: `http` (default) or `influxdb_table`
-- `feed`: USGS feed key when using HTTP mode
-- `source_url`: custom URL when using HTTP mode
-- `source_format`: `usgs_geojson` or `flat_json`
-- `source_table`: source table for `influxdb_table` mode (default `quake`)
-- `source_query`: optional SQL override for table mode
-- `lookback_minutes`: table query lookback window (default `15`)
-- `measurement`: destination measurement (default `earthquakes`)
-- `write_quake_schema`: write only the canonical `quake` columns (`depth`, `depthError`, `dmin`, `gap`, `horizontalError`, `id`, `latitude`, `locationSource`, `longitude`, `mag`, `magError`, `magNst`, `magSource`, `magType`, `net`, `nst`, `place`, `rms`, `status`, `time`, and `type`). Use with `measurement=quake`. Note: `depthError`, `horizontalError`, `magError`, `magNst`, `locationSource`, and `magSource` exist only in USGS CSV feeds, not GeoJSON, so they are written only when a `flat_json` or `influxdb_table` source supplies them.
+This plugin includes a JSON metadata schema in its docstring that defines supported trigger types and configuration parameters. This metadata enables the [InfluxDB 3 Explorer](https://docs.influxdata.com/influxdb3/explorer/) UI to display and configure the plugin.
 
-- `min_magnitude`: optional minimum magnitude filter; when omitted, nothing is filtered (USGS feeds include negative-magnitude microseisms)
-- `max_events`: max events processed per run
-- `skip_unchanged`: skip unchanged events based on cached marker
-- `use_event_timestamp`: write event timestamp vs trigger time
+### Data selection parameters
 
-## Install / deploy
+| Parameter          | Type    | Default        | Description                                                                                                                                                  |
+|--------------------|---------|----------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `feed`             | string  | all_hour       | USGS GeoJSON feed key: `all`, `significant`, `4.5`, `2.5`, or `1.0` combined with `_hour`, `_day`, `_week`, or `_month` (for example `significant_day`)       |
+| `source_type`      | string  | http           | Data source type: `http` fetches JSON from `source_url` or `feed`; `influxdb_table` queries an existing table in the trigger database                        |
+| `source_url`       | string  | none           | Custom source URL (`http` or `https` only). When provided, overrides `feed` and uses `source_format` parsing                                                 |
+| `source_format`    | string  | usgs_geojson   | Source parser for HTTP mode: `usgs_geojson` or `flat_json` (for records like `{id, latitude, longitude, mag, time, ...}`)                                    |
+| `source_table`     | string  | quake          | Source table name when `source_type=influxdb_table`                                                                                                          |
+| `source_query`     | string  | none           | Optional SQL override for `influxdb_table` mode; disables watermark paging                                                                                    |
+| `lookback_minutes` | integer | 15             | Initial lookback window for `influxdb_table` mode. Later runs page forward from the cached fetch watermark while `skip_unchanged=true`                        |
 
-### Option A: Generic InfluxDB 3 trigger create (from a node with `influxdb3` CLI)
+### Optional parameters
 
-1) Place plugin file where CLI can read it:
+| Parameter             | Type    | Default                         | Description                                                                                                                                                  |
+|-----------------------|---------|---------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `measurement`         | string  | earthquakes                     | Destination measurement name                                                                                                                                 |
+| `write_quake_schema`  | boolean | false                           | Write events using the canonical `quake` table's column names; no tags or normalized-only columns are written. Use with `measurement=quake`                  |
+| `min_magnitude`       | float   | none                            | Optional minimum magnitude. When omitted, nothing is filtered (USGS feeds include negative-magnitude microseisms and events without a magnitude)             |
+| `max_events`          | integer | 250                             | Maximum events written per run, applied after filtering. Events deferred by the cap remain uncached and are written on a later run                            |
+| `use_event_timestamp` | boolean | true                            | Use the event time as the point timestamp; `false` uses the trigger execution time. Ignored (forced `true`) when `write_quake_schema=true`                   |
+| `skip_unchanged`      | boolean | true                            | Skip events whose update marker is not newer than the last written copy of the same event. Events without an id are always written                            |
+| `user_agent`          | string  | InfluxDB3-Earthquake-Plugin/1.0 | Custom User-Agent header for API requests                                                                                                                    |
+| `enable_full_logging` | boolean | false                           | When `true`, full exception messages are logged. When `false` (default), only exception types are logged                                                     |
 
-- `earthquake_sampler.py`
+## Schema requirements
 
-2) Create a scheduled trigger that fetches the USGS GeoJSON `all_hour` feed and writes it directly to `usgs.quake`:
+`write_quake_schema=true` targets an existing `quake` table with the USGS CSV column layout (`depth`, `dmin`, `gap`, `id`, `latitude`, `longitude`, `mag`, `magType`, `net`, `nst`, `place`, `rms`, `status`, `time`, `type`, and the CSV-only columns below).
+
+- All numeric columns (including `nst` and `magNst`) are written as float64. This matches quake tables created by CSV import, where numeric columns containing blanks are inferred as doubles. If your existing table stores these columns as int64, the writes are rejected with a type conflict.
+- `depthError`, `horizontalError`, `magError`, `magNst`, `locationSource`, and `magSource` exist only in USGS CSV feeds, not GeoJSON, so they are written only when a `flat_json` or `influxdb_table` source supplies them.
+- Quake-schema rows carry no tags, so point identity rests entirely on the timestamp. USGS supplies millisecond-precision times; the plugin fills the unused sub-millisecond bits with a stable per-event offset so two earthquakes in the same millisecond do not overwrite each other. Millisecond-level time is unchanged.
+
+## Software requirements
+
+- **InfluxDB 3 Core/Enterprise**: 3.8.2 or later (the plugin writes with `write_sync`), with the Processing Engine enabled
+- **Python packages**: No additional packages required (uses the Python standard library only)
+- **Network access**: Outbound HTTPS access to `earthquake.usgs.gov` (HTTP mode)
+
+### Installation steps
+
+1. Start InfluxDB 3 with the Processing Engine enabled (`--plugin-dir /path/to/plugins`):
+
+   ```bash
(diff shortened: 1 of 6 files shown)


caterryan:
Thank you @rkinginflux! We are reviewing this now

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+            "required": false
+        },
+        {
+            "name": "source_table",
+            "example": "quake",
+            "description": "Source table name when `source_type=influxdb_table`. Defaults to quake.",
+            "required": false
+        },
+        {
+            "name": "source_query",
```
Aliaksei-Kharlap, line 36: The metadata is invalid due to this example. Parsing with `json.loads()` throws an error, preventing the Explorer UI from correctly displaying the plugin information

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+        "mmi": item.get("mmi"),
+        "nst": item.get("nst"),
+        "depth_km": item.get("depth"),
+        "longitude": item.get("longitude"),
+        "latitude": item.get("latitude"),
+        "gap_degrees": item.get("gap"),
+        "distance_km": item.get("dmin"),
+        "rms": item.get("rms"),
+        "updated_ms": item.get("updated") or item.get("updatedMs"),
+        "event_time_ns": _to_ns_from_iso(item.get("time")) or _to_ns_from_ms(item.get("timeMs")),
```
Aliaksei-Kharlap: The `source_type=influxdb_table` mode always fails with an error because the InfluxDB query returns the `time` column as an integer rather than a datetime string

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+                    measurement=measurement,
+                    event=event,
+                    fallback_ts_ns=now_ns,
+                    use_event_timestamp=use_event_timestamp,
+                )
+            if did_write:
+                written += 1
+                if max_marker_this_run is None or marker > max_marker_this_run:
+                    max_marker_this_run = marker
+        except Exception as e:
```
Aliaksei-Kharlap, line 901: The legacy `influxdb3_local.write(line)` method is currently used for writing; this method does not raise database errors. Records accumulate in a buffer, and the actual write operation begins only after the plugin finishes execution. It is recommended to use `influxdb3_local.write_sync(line, no_sync=True)` instead

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+            skipped += 1
+            continue
+
+        marker = _to_update_marker_ms(event)
+        if skip_unchanged and last_seen_marker is not None and marker <= last_seen_marker:
+            skipped += 1
+            continue
+
+        try:
+            if write_quake_schema:
```
Aliaksei-Kharlap, line 881: `write_quake_schema=true` does not write a single tag.
The series key is empty, so data point uniqueness relies solely on the timestamp. USGS provides time with millisecond precision—collisions between different earthquakes are a real possibility and result in data loss. And
`use_event_timestamp=false` in this mode guarantees exactly one row per run.

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+        last_seen_marker = None
+
+    fetched = 0
+    written = 0
+    skipped = 0
+    max_marker_this_run: Optional[int] = last_seen_marker
+
+    normalized.sort(key=_to_update_marker_ms, reverse=True)
+
+    for event in normalized:
```
Aliaksei-Kharlap, line 856: `max_events` is applied *before* filters, not after.
`fetched` is incremented for every event processed, including those discarded due to `min_magnitude` and `skip_unchanged`. Metadata claims it applies "after filtering and sorting"

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+    timestamp_ns = fallback_ts_ns
+    event_time_ns = event.get("event_time_ns")
+    if use_event_timestamp and event_time_ns is not None:
+        try:
+            timestamp_ns = int(event_time_ns)
+        except (TypeError, ValueError):
+            pass
+
+    line = _line_builder(measurement).time_ns(timestamp_ns)
+    line.tag("event_id", _safe_tag(event.get("event_id"), "unknown"))
```
Aliaksei-Kharlap: Using `event_id` as a tag will significantly increase cardinality, it is recommended to store such values ​​as a string field.

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+    """Write a normalized USGS event to a table using only the canonical quake schema."""
+    timestamp_ns = fallback_ts_ns
+    if use_event_timestamp and event.get("event_time_ns") is not None:
+        try:
+            timestamp_ns = int(event["event_time_ns"])
+        except (TypeError, ValueError):
+            pass
+
+    line = _line_builder(measurement).time_ns(timestamp_ns)
+    for column, event_key in QUAKE_FLOAT_COLUMN_MAP.items():
```
Aliaksei-Kharlap: `nst` and `magNst` are integer counters, yet they are defined as `float64_field`. The docstring promises "the existing quake table's column names and types"; if the existing table has `int64` columns, the record will be rejected due to a type conflict. However, in `_write_event`, these same fields are written as `int64`—an internal inconsistency

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+            if math.isfinite(val):
+                line.float64_field(field_name, val)
+        except (TypeError, ValueError):
+            continue
+
+    for field_name in (
+        "significance",
+        "felt_reports",
+        "tsunami",
+        "mmi",
```
Aliaksei-Kharlap: mmi is written as int64_field, but MMI is a fractional value, and int() truncates it

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+        except (TypeError, ValueError):
+            pass
+    return 0
+
+
+def _fetch_payload(url: str, user_agent: str) -> Dict[str, Any]:
+    req = Request(url)
+    req.add_header("User-Agent", user_agent)
+    req.add_header("Accept", "application/geo+json, application/json")
+    with urlopen(req, timeout=20) as response:
```
Aliaksei-Kharlap, line 402: `source_url` without schema validation—`urlopen` supports `file://` and `ftp://`, meaning `source_url=file:///…` allows the database process to read local files

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+    max_events: int,
+    lookback_minutes: int,
+    source_query: str,
+) -> List[Dict[str, Any]]:
+    if source_query.strip():
+        query = source_query
+    else:
+        safe_lookback = max(1, int(lookback_minutes))
+        query = (
+            f'SELECT * FROM "{source_table}" '
```
Aliaksei-Kharlap: source_table is inserted into the SQL without escaping

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+    source_query = _safe_string(args.get("source_query"))
+    lookback_minutes = max(1, _parse_int(args.get("lookback_minutes"), 15))
+
+    source = source_url if source_url else feed
+    if source_type == "influxdb_table":
+        source = source_query if source_query else source_table
+
+    measurement = _safe_string(args.get("measurement", "earthquakes")) or "earthquakes"
+    write_quake_schema = _parse_bool(args.get("write_quake_schema"), False)
+    min_magnitude = _parse_float(args.get("min_magnitude"), 0.0)
```
Aliaksei-Kharlap: The default `min_magnitude=0.0` silently cuts 3–4% of the feed. 
Measurements: `all_day` 10/299, `all_week` 68/2336, `all_month` 486/11105 — these are all `type=earthquake` with negative magnitudes (down to -1.22), representing normal microseismic activity. "Defaults to 0" is interpreted
as "do not filter"; to actually avoid filtering, one has to explicitly pass a negative number. The default should be `None`

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,648 @@
...
+        return []
+    return [r for r in rows if isinstance(r, dict)]
+
+
+def _normalize_usgs_feature(feature: Dict[str, Any]) -> Dict[str, Any]:
+    properties = feature.get("properties", {}) if isinstance(feature, dict) else {}
+    geometry = feature.get("geometry", {}) if isinstance(feature, dict) else {}
+    coordinates = geometry.get("coordinates", []) if isinstance(geometry, dict) else []
+
+    return {
```
Aliaksei-Kharlap, line 452: The keys `depthError`, `horizontalError`, `magError`, `magNst`, `locationSource`, and `magSource` are column names from the USGS CSV feed; they do not exist in GeoJSON

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Intentional, now documented (841e657): quake mode targets CSV-imported tables, where numeric columns containing blanks are inferred as float64 — and `magNst` never appears in GeoJSON anyway. The README/metadata now state the float64 assumption explicitly.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,717 @@
...
+            magnitude = float(mag) if mag is not None else None
+        except (TypeError, ValueError):
+            magnitude = None
+
+        if min_magnitude is not None and (magnitude is None or magnitude < min_magnitude):
+            skipped += 1
+            continue
+
+        marker = _to_update_marker_ms(event)
+        if skip_unchanged and last_seen_marker is not None and marker <= last_seen_marker:
```
caterryan: The single watermark loses events: with markers 300/200/100 and `max_events=2`, run 1 writes 300+200 and caches 300; run 2 skips all three, so 100 is never written. The same mechanism permanently blocks previously filtered events if `min_magnitude` is later lowered. Fix planned: per-event marker cache (`event_id` → marker, TTL-bounded) instead of one global watermark.

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,717 @@
...
+    if source_query.strip():
+        query = source_query
+    else:
+        safe_lookback = max(1, int(lookback_minutes))
+        safe_table = source_table.replace('"', '""')
+        query = (
+            f'SELECT * FROM "{safe_table}" '
+            f"WHERE time >= now() - INTERVAL '{safe_lookback} minutes' "
+            f"ORDER BY time DESC "
+            f"LIMIT {max_events}"
```
caterryan, line 439: `ORDER BY time DESC LIMIT` fetches only the newest rows in the window, so under load rows between the watermark and the newest N are never fetched — and are skipped once the watermark advances past them. Fix planned: page ascending from the watermark (`WHERE time > $watermark ORDER BY time ASC LIMIT N`).

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,717 @@
...
+    call_time: datetime,
+    args: Optional[Dict[str, Any]] = None,
+) -> None:
+    task_id = str(uuid.uuid4())
+
+    global _ENABLE_FULL_LOGGING
+    _ENABLE_FULL_LOGGING = _parse_bool((args or {}).get("enable_full_logging"), False)
+
+    args = args or {}
+    source_type = _safe_string(args.get("source_type", "http")).lower() or "http"
```
caterryan: Config parsing silently falls back on invalid input: bad `source_type` becomes `http`, bad `feed` becomes `all_hour`, unparseable `min_magnitude` becomes no-filter, and unrecognized boolean strings become `false` (`skip_unchanged=ture` would silently disable dedup). Fix planned: error on invalid enums and unparseable numbers/booleans.

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,717 @@
...
+
+    feed = _safe_string(args.get("feed", "all_hour"))
+    if feed not in FEED_URLS:
+        feed = "all_hour"
+
+    source_table = _safe_string(args.get("source_table", "quake")) or "quake"
+    source_query = _safe_string(args.get("source_query"))
+    lookback_minutes = max(1, _parse_int(args.get("lookback_minutes"), 15))
+
+    source = source_url if source_url else feed
```
caterryan: `source` carries the full custom URL / `source_query` into info/error logs regardless of `enable_full_logging`, so URL credentials/query params or SQL literals can leak. Fix planned: strip userinfo/query-string from logged URLs and truncate `source_query`.

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,717 @@
...
+import uuid
+import zlib
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
+from urllib.error import HTTPError, URLError
+from urllib.parse import urlparse
+from urllib.request import Request, urlopen
+
+
+def _line_builder(measurement: str):
```
caterryan: The builtins-only lookup fails if the engine injects `LineBuilder` into module globals instead. Fix planned: use bare `LineBuilder(...)` (`# noqa: F821`) like the other plugins — name resolution covers both globals and builtins.

influxdata/earthquake_sampler/earthquake_sampler.py
```diff
@@ -0,0 +1,717 @@
...
+        value = event.get(event_key)
+        if value is not None:
+            line.string_field(column, _safe_string(value))
+
+    influxdb3_local.write_sync(line, no_sync=True)
+    return True
+
+
+
+def process_scheduled_call(
```
caterryan, line 724: No unit tests for a two-mode/two-schema ingestion plugin; comparable plugins (signal_generator, nori_regression, valuecounter) ship mock-based tests. Fix planned: `test_earthquake_sampler.py` with stubbed `LineBuilder`/`influxdb3_local` covering both source modes, rerun dedup, filtering, and timestamps.

influxdata/earthquake_sampler/README.md
```diff
@@ -0,0 +1,112 @@
+# InfluxDB 3 Earthquake Sampler Plugin
```
caterryan, line 1: `scripts/validate_readme.py` fails this README with 12 errors / 8 warnings: missing emoji metadata line, required sections (Description/Configuration/Trigger setup/Example usage/Code overview/Troubleshooting/Questions-Comments), parameter tables, and installation steps; log queries must use `ORDER BY event_time` (the `time` column doesn't exist on `system.processing_engine_logs`). Fix planned: rewrite to README_TEMPLATE.md structure modeled on nws_weather, keeping the quake-schema notes in parameter docs/Troubleshooting.

caterryan (commented):
Claude found some additional items

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Fixed.

influxdata/earthquake_sampler/README.md
caterryan: Fixed.

influxdata/earthquake_sampler/earthquake_sampler.py
caterryan: Deferred — tests are not being added in this PR.

caterryan (approved):

PR #154, opened 2026-09-01
nori_regression: correct the plugin's Synthefy-facing details (key env var, model list, required model)

ajanbekzat:
Three corrections to the plugin's Synthefy-facing details, one per commit, so any of them can be dropped.

## 1. Rename the API key environment variable to `SYNTHEFY_NORI_API_KEY`

An InfluxDB host runs plugins from several authors and shares one environment. An unprefixed `NORI_API_KEY` does not say whose service it authenticates against, and it is the kind of name that collides. The vendor prefix makes it unambiguous and matches how operators receive the key, from the Synthefy console.

The `X-Nori-Api-Key` header is unchanged, and so is the resolution order: a non-empty header wins (HTTP trigger), otherwise the environment variable.

**No compatibility shim.** The plugin is `[BETA]` in the registry with no known deployments, and a missing key is already a loud failure that names the variable to set:

```
No Nori API key. Set the SYNTHEFY_NORI_API_KEY environment variable on the InfluxDB host,
or pass a 'X-Nori-Api-Key: <key>' header (HTTP trigger).
```

So an operator who upgrades with only the old variable set is told exactly what to change. Happy to add a transitional fallback if you would rather not break even an unknown user.

## 2. Point the model list at Synthefy's docs instead of duplicating it

The plugin kept its own list of model slugs in four places: the README's "Supported models" table, the `model` argument's help text, the retired-slug error, and the commented TOML config. All four listed `synthefy/nori-30m` and `synthefy/nori-6m` only, while Synthefy currently publishes four entries including `synthefy/nori-100m`. A reader of this README would not know the larger model exists.

Which models exist is not something this plugin can keep correct: it changes when Synthefy releases a variant, with no signal to this repository. Each copy is now a pointer to [docs.synthefy.com/nori/quickstart#models](https://docs.synthefy.com/nori/quickstart#models). The plugin still names its own default, since that is the plugin's behaviour rather than the vendor's catalogue.

The per-slug cold-start figures ("about 69 seconds for nori-6m and 125 seconds for nori-30m") become a range in the same change. They were measured against two specific variants, and a cold start has since been observed at close to four minutes, so a precise pair of numbers reads as more of a guarantee than it is. The `request_timeout` default of `300s` is unchanged and still absorbs it.

## 3. Require an explicit `model`

The plugin defaulted `model` to `synthefy/nori-30m`. That slug selects a **priced** model, so defaulting it spends the operator's money on a variant they never named, and pins them to a slug that can be retired from under them. That has already happened once: the bare `synthefy/nori` slug was retired while it was this plugin's default.

Synthefy's own client and local package both refuse to default a model and raise instead, for exactly this reason. The plugin now matches them. `model` joins `measurement`, `field` and `feature_fields` as required, and the error names the published list:

```
`model` is required: the Nori gateway slug to call, for example 'synthefy/nori-30m'.
There is no default, because the slug selects a priced model. The current slugs are
listed at https://docs.synthefy.com/nori/quickstart#models.
```

The shipped TOML template now sets `model` rather than commenting it out, and the README moves it to the required-parameter table.

## Testing

Unit tests: **194 passed**, up from 190 on `main`. The four added tests pin the environment variable name and cover the `model` requirement, both of which are public contract because the error messages instruct operators about them.

End to end against InfluxDB 3 Core with the modified plugin mounted, a 72-row table whose last 12 rows are missing the target, and the **production** Nori gateway with a production key and `model=synthefy/nori-30m`:

| Scenario | Result |
|---|---|
| `SYNTHEFY_NORI_API_KEY` set, `model` named | `{"status": "success", "result": {"written": 12}}` |
| no `model` argument | fails: `` `model` is required `` |
| `model` set in the request body | refused: operator settings cannot come from the body |
| only the old `NORI_API_KEY` set | fails, message names `SYNTHEFY_NORI_API_KEY` |
| neither set, no header | fails, message names `SYNTHEFY_NORI_API_KEY` |
| no env var, key in `X-Nori-Api-Key` | `{"status": "success", "result": {"written": 12}}` |

The README also passes the repository's own validator (`scripts/validate_readme.py --plugins nori_regression`): 0 errors, 0 warnings.

The retired-slug error was rendered to confirm the new interpolation:

```
model slug 'synthefy/nori' is retired and no longer routes. Use 'synthefy/nori-30m' or
another current slug, listed at https://docs.synthefy.com/nori/quickstart#models.
```

## Note

`NORI_GATEWAY_URL`, the optional private-gateway override, keeps its current name here to keep the change reviewable. Say the word if you would prefer both renamed together.

### DIFF
--- a/influxdata/nori_regression/README.md
+++ b/influxdata/nori_regression/README.md
@@ -107,7 +107,7 @@ because every string form splits on whitespace. Name such a column from a TOML a
 | `max_predict_rows` | integer | `5000` | Cap on rows predicted per run; the most recent rows are kept and the rest wait for a later run. |
 | `max_read_rows` | integer | `50000` | Ceiling on rows read from InfluxDB in one run, applied as a `LIMIT` on the query. The most recent rows are read, and a truncated read is logged with a warning. This bounds the plugin's memory: a row costs roughly 0.7 KB while it is held, so the default is about 35 MB. |
 | `predict_batch_size` | integer | `1000` | Rows per gateway call. Each batch re-sends the training context and is billed separately, so a larger value costs less. |
-| `request_timeout` | string | `300s` | Timeout for one gateway call. A cold start has been measured at 60-130 seconds, so keep this well above that. |
+| `request_timeout` | string | `300s` | Timeout for one gateway call. A model that has scaled to zero cold-starts on the first request, measured between roughly one and four minutes depending on the variant, so keep this well above the warm response time. |
 | `max_retries` | integer | `3` | Maximum attempts per gateway call and per write. `1` disables retry. |
 | `config_file_path` | string | *(none)* | Path to a TOML file supplying every parameter, relative to `PLUGIN_DIR`. Cannot be combined with other inline arguments or a request body. |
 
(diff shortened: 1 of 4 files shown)


caterryan (approved):

ajanbekzat:
thanks @caterryan , is it possible to get it merged soon? It's blocking some tasks

caterryan:
Thanks for the contribution!

PR #155, opened 2026-09-01
nori_regression: point the model list at Synthefy's docs instead of duplicating it

ajanbekzat:
## What

The plugin kept its own list of Nori model slugs in four places: the README's "Supported models" table, the `model` argument's help text, the retired-slug error message, and the commented TOML config. Each is now a pointer to Synthefy's published list at [docs.synthefy.com/nori/quickstart#models](https://docs.synthefy.com/nori/quickstart#models).

The plugin still names its own default (`synthefy/nori-30m`), because that is the plugin's behaviour rather than the vendor's catalogue.

## Why

The copies were already stale. They list `synthefy/nori-30m` and `synthefy/nori-6m` only, while Synthefy currently publishes four entries, including `synthefy/nori-100m` and an API-only `synthefy/nori-30m-thinking-medium`. A user reading this README would not know the larger model exists.

Which models exist is not something this plugin can keep correct: it changes when Synthefy releases a variant, with no signal to this repository. Pointing at the vendor's list removes that failure mode instead of refreshing it once.

## Also in this change

The per-slug cold-start figures ("about 69 seconds for nori-6m and 125 seconds for nori-30m") are replaced by a range. They were measured against two specific variants, and a cold start has since been observed at close to four minutes, so a precise pair of numbers reads as more of a guarantee than it is. The `request_timeout` default of `300s` is unchanged and still absorbs it.

## Testing

```
pytest influxdata/nori_regression/test_nori_regression.py
190 passed
```

The retired-slug error was rendered to confirm the new interpolation:

```
model slug 'synthefy/nori' is retired and no longer routes. Use 'synthefy/nori-30m' or
another current slug, listed at https://docs.synthefy.com/nori/quickstart#models.
```

## Note

This is independent of #154, which renames the plugin's API key environment variable. Both touch `README.md` but in different sections, so they can merge in either order.

### DIFF
--- a/influxdata/nori_regression/README.md
+++ b/influxdata/nori_regression/README.md
@@ -107,7 +107,7 @@ because every string form splits on whitespace. Name such a column from a TOML a
 | `max_predict_rows` | integer | `5000` | Cap on rows predicted per run; the most recent rows are kept and the rest wait for a later run. |
 | `max_read_rows` | integer | `50000` | Ceiling on rows read from InfluxDB in one run, applied as a `LIMIT` on the query. The most recent rows are read, and a truncated read is logged with a warning. This bounds the plugin's memory: a row costs roughly 0.7 KB while it is held, so the default is about 35 MB. |
 | `predict_batch_size` | integer | `1000` | Rows per gateway call. Each batch re-sends the training context and is billed separately, so a larger value costs less. |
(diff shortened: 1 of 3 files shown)


ajanbekzat:
Folded into #154 so the plugin's Synthefy-facing details land in one change. Closing this one.

PR #162, opened 2026-09-11
fix(mqtt_subscriber): drain the tail before disconnecting

### LINKED ISSUE #161: mqtt_subscriber: messages received between the drain and the disconnect are acknowledged and then discarded
## Summary

`mqtt_subscriber` loses messages that arrive between its queue drain and its disconnect. The
message is acknowledged to the broker by paho, then discarded with the client object, and QoS 1
never re-delivers it. On a scheduled trigger the lost window is roughly half a second of every
cycle, so a steady low-rate stream loses a few percent of its messages, silently.

Measured on a `every:10s` trigger against Mosquitto: 13 of about 250 messages were lost over
four weeks. Every lost message was published between 0.55 s and 1.10 s after a cycle boundary.
Nothing published outside that window was lost.

## Why the window exists

In `process_scheduled_call`:

1. The client connects and subscribes (about T+0.06 s).
2. `time.sleep(0.5)`.
3. `messages = mqtt_client.get_messages()` drains the queue (about T+0.56 s).
4. Parse, then `influxdb3_local.write_sync(...)`.
5. `finally: mqtt_client.disconnect()` (about T+1.06 s).

The network loop keeps running through steps 4 and 5. A PUBLISH that arrives there is handled by
`_on_message`, which puts it on `self.message_queue` — after the only drain. paho has already sent
the PUBACK, so the broker considers it delivered and drops it from the persistent session queue.
The client is then disconnected and the queue is garbage collected with the message still in it.

Messages that arrive while the plugin is *offline* are safe, because `clean_session=False` keeps
them queued at the broker. The loss is specific to the plugin's own connected tail.

## Reproduce

1. Create a scheduled `mqtt_subscriber` trigger, `every:10s`, QoS 1, against any broker.
2. Publish one message per cycle at a fixed offset of 0.75 s past each 10-second boundary.
3. Compare the published count with the row count written to the target table.

Every message is acknowledged by the broker and never appears in the table.

## Suggested fix

Stop receiving before the parse/write phase, then drain once more. `disconnect()` already calls
`loop_stop()` first, so once it returns no further message can be accepted or acknowledged, and
anything the broker still holds is re-delivered on the next cycle.

```python
        # Retrieve messages from queue
        messages: list = mqtt_client.get_messages()

+       # Stop receiving before the parse/write phase, then drain once more, so a
+       # message accepted in the tail is not acknowledged and then discarded.
+       mqtt_client.disconnect()
+       messages.extend(mqtt_client.get_messages())
```

`disconnect()` needs an idempotence guard, because the `finally` block calls it again:

```python
     def disconnect(self):
         """Disconnect from MQTT broker"""
-        if self.client:
+        if self.client and not self.disconnected:
+            self.disconnected = True
             self.client.loop_stop()
             self.client.disconnect()
```

with `self.disconnected: bool = False` initialised next to `self.connected`.

Applied to a live deployment, 13 messages republished at 0.75 s past the boundary — the middle of
the old loss window — all landed. I am happy to open a pull request if that helps.

## Environment

- InfluxDB 3 Core 3.11.4
- Plugin: `gh:influxdata/mqtt_subscriber/mqtt_subscriber.py` (`main`)
- Broker: Mosquitto, TLS, QoS 1, `clean_session=False`, explicit `client_id`

Sawtaytoes:
Fixes #161.

`mqtt_subscriber` drains its queue about 0.56 s into each cycle, then parses and writes, then disconnects about 0.5 s later. The network loop runs throughout, so a PUBLISH arriving in that tail is handled by `_on_message` and queued after the only drain. paho has already sent the PUBACK, so the broker drops it from the persistent session, and the client is then disconnected with the message still in its queue.

Measured on a live `every:10s` trigger: 13 of about 250 messages lost over four weeks, every one of them published between 0.55 s and 1.10 s after a cycle boundary, and nothing lost outside it.

This change stops receiving before the parse/write phase and drains once more. `disconnect()` already calls `loop_stop()` first, so once it returns nothing further can be accepted or acknowledged, and whatever the broker still holds is re-delivered on the next cycle. `disconnect()` gains an idempotence guard because the `finally` block calls it again.

Verified on the same deployment: 13 messages republished at 0.75 s past the boundary — the middle of the old loss window — all landed.


influxdata/mqtt_subscriber/mqtt_subscriber.py
```diff
@@ -938,6 +938,7 @@ def __init__(self, config: dict[str, Any], influxdb3_local, task_id):
         self.client = None
         self.message_queue: Queue = Queue()
         self.connected: bool = False
+        self.disconnected: bool = False
```
caterryan: This flag can go away if `disconnect()` clears `self.client` instead (see the suggestion on `disconnect()` below). Two overlapping booleans (`connected` / `disconnected`) that can both be `False` at once are easy to misread.

```suggestion
```

influxdata/mqtt_subscriber/mqtt_subscriber.py
```diff
@@ -1206,7 +1207,8 @@ def get_drop_counters(self) -> dict[str, int]:
 
     def disconnect(self):
         """Disconnect from MQTT broker"""
-        if self.client:
+        if self.client and not self.disconnected:
+            self.disconnected = True
             self.client.loop_stop()
             self.client.disconnect()
             self.connected = False
```
caterryan: Suggest making idempotence fall out of the existing `self.client` guard by moving the client into a local and clearing the attribute, rather than adding a second flag. While this method is being touched: today a failed `connect()` still logs "Disconnected from MQTT broker" from the `finally` block, which is misleading. Snapshotting `connected` before `client.disconnect()` (paho's own on_disconnect callback clears it synchronously during that call) lets the existing log line below fire only when a session actually existed.

```suggestion
        if self.client:
            client, self.client = self.client, None
            was_connected: bool = self.connected
            client.loop_stop()
            client.disconnect()
            self.connected = False
            if not was_connected:
                return
```

influxdata/mqtt_subscriber/mqtt_subscriber.py
```diff
@@ -2094,6 +2096,15 @@ def process_scheduled_call(
...
         messages: list = mqtt_client.get_messages()
 
+        # Stop receiving before the parse/write phase, then drain once more.
+        # A message that arrives after the drain above is PUBACKed by paho and then
+        # thrown away with the client, and QoS 1 never re-delivers it, so every
+        # publish landing in that tail window is lost. Disconnecting first makes the
+        # window microseconds wide, and the second drain collects whatever fell into
+        # it. Anything the broker still holds is re-delivered on the next cycle.
+        mqtt_client.disconnect()
+        messages.extend(mqtt_client.get_messages())
```
caterryan: With `disconnect()` moved up, the first drain is redundant and the comment undersells the fix. `disconnect()` calls `loop_stop()`, which joins paho's network thread, so once it returns there is no producer left and a single `get_messages()` afterwards is complete. There is no "microseconds wide" window to collect from. Collapsing to one drain makes that guarantee obvious to the next reader.

I confirmed the join semantics in paho 2.1.0 (`loop_stop` -> `Thread.join`; `_handle_publish` runs `on_message` before `_send_puback` for QoS 1) and reproduced both the bug and the fix against Mosquitto: see the review summary.

```suggestion
        # Stop receiving before the parse/write phase. disconnect() joins the
        # network thread, so once it returns no further on_message can run and a
        # single drain here is complete. Before this change, a message that arrived
        # after the drain was PUBACKed by paho and then thrown away with the client,
        # and QoS 1 never re-delivers it. Anything the broker still holds at the
        # moment we stop reading is re-delivered on the next cycle.
        mqtt_client.disconnect()
        messages: list = mqtt_client.get_messages()
```

caterryan (changes requested):
I was able to repro this. Two changes and then this is ready:

1. `influxdata/mqtt_subscriber/manifest.toml`: bump `version` to `0.2.1`
2. `influxdata/mqtt_subscriber/README.md` line 9 currently says persistent sessions "ensure message delivery". Please add a sentence that delivery is at-least-once: a message can be written twice if the broker misses an acknowledgement at disconnect, and mapping a timestamp field from the payload makes re-deliveries idempotent.

Sawtaytoes:
Addressed both review items in 9ce7fe4:

- bumped the plugin version to `0.2.1`
- documented at-least-once delivery, the possible duplicate write at disconnect, and payload `timestamp_field` mapping for idempotent redelivery

I also ran `influxdb3-plugin validate influxdata/mqtt_subscriber --output json` locally, and it returned `{"status":"ok","result":{}}`.

caterryan (approved):

caterryan:
Thank you for the contribution, @Sawtaytoes!

PR #163, opened 2026-09-14
Feat/update utils package

Aliaksei-Kharlap:
Reworks how a plugin loads its configuration

A plugin composes the layers it needs and validates them once:

  ```python
  cfg = load_config(
      parse_env(ENV),
      parse_trigger_args(args),
      parse_toml(args.get("config_file_path")),
      parse_json_body(request_body, BODY),
      parse_request_headers(request_headers, HEADERS),
      parse_query_parameters(query_parameters, QUERY),
      validators=VALIDATORS,
  )

  - sources — one parser per place configuration comes from, all with the
    same shape. Trigger arguments, the TOML file and the environment are no
    longer handled differently from the request body, headers and query string.
  - KeySpec(allowlist, denylist, rename, unknown) — says which keys of a
    source contribute and under what names. unknown="reject" names a refused
    key back to whoever sent it; parse_env requires an allowlist, since the
    process environment belongs to the host.
  - load_config(*layers, validators=...) — the argument order is the
    precedence, lowest first. Any number of layers, in any order, including
    dicts the plugin builds itself.
  - validation — Validator and validate() are now the package's own,
    so dynaconf is gone and the package has no dependencies.
  - load_plugin_config stays supported for the plugins already on it;
    load_config is the one to reach for in new code.


influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,257 @@
...
+from ._utils import is_blank
+
+__all__ = ["Validator", "validate"]
+
+_MISSING = object()
+
+
+def _show(value) -> str:
+    """A value as it reads in a message: strings quoted, everything else plain."""
+    return repr(value) if isinstance(value, (str, bytes)) else str(value)
```
caterryan, line 23: **Bound the value in error messages.** A value that fails a check is copied whole into the message: a 200 KB body value produces a 200 KB `ValueError`, which then lands in the HTTP 400 response and in the log line. With the 10 MB body limit, one request can produce a 20 MB response.

`_select` already caps refused keys at 64 characters. Do the same here: cut the value to ~200 characters with an ellipsis, and route the two `{value!r}` sites below (cast failure, condition failure) through `_show` as well.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,257 @@
...
+
+        if self.condition is not None:
+            try:
+                allowed = self.condition(value)
+            except Exception as exc:
+                raise ValueError(
+                    f"{name} cannot be checked with condition: {exc}"
+                ) from exc
+            if not allowed:
+                raise ValueError(f"{name} is not allowed: {value!r}")
```
caterryan, line 234: Both `{value!r}` sites here should go through the bounded `_show` (see the comment on `_show` above). Otherwise a cast or condition failure still echoes the full value.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,257 @@
...
+    each use, so one rule's default cannot be changed through the values it
+    fills in. Checks are named explicitly, so a misspelled one is a
+    ``TypeError`` where the rule is written.
+    """
+
+    def __init__(
+        self,
+        *names: str,
+        required: bool | None = None,
+        must_exist: bool | None = None,
```
caterryan: **Drop `must_exist`; `required` is the one keyword.** In dynaconf `must_exist=False` meant "must not be set". Here it is silently no rule at all, so a rule ported from dynaconf lets a value through without a word.

Rather than carry an alias with that trap in it, remove the parameter and the `bool(required or must_exist)` on line 141. Then:
- delete `test_must_exist_is_the_same_rule` in `tests/test_validation.py`;
- switch the two `must_exist=True` rules in `influxdata/geo_enrichment/geo_enrichment.py:1207-1208` to `required=True` in this PR, since that plugin pins `>=0.4.0`.

This is a breaking change and that is accepted.

influxdata-plugin-utils/src/influxdata_plugin_utils/sources.py
```diff
@@ -0,0 +1,489 @@
...
+
+
+def is_toml_path(path) -> bool:
+    """Does this path name a TOML file? What to do when it does not is the caller's call."""
+    return as_text(path).strip().lower().endswith(".toml")
+
+
+def header_key(name) -> str:
+    """Turn a protocol-level name into a config key: ``X-Api-Key`` -> ``x_api_key``."""
+    return as_text(name).strip().lower().replace("-", "_")
```
caterryan: **Lowercase only; keep hyphens.** Header names should be matched case-insensitively but no longer rewritten from `-` to `_`. A plugin that wants `api_key` says so with `rename`. So this becomes `.strip().lower()`, `X-Api-Key` lands as `x-api-key`, and `X-Api-Key` vs `x_api_key` are simply two different keys.

Update with it:
- this docstring and the `parse_request_headers` docstring (lines 429-432, "casing and hyphenation");
- the "two spellings" wording in `_folded_rename`;
- the tests in `tests/test_sources.py` that assert the underscore form (lines 238-240, 245-249, 277-284).

influxdata-plugin-utils/src/influxdata_plugin_utils/sources.py
```diff
@@ -0,0 +1,489 @@
...
+    it, so a token needs a header of your own.
+
+    Args:
+        request_headers: Headers as delivered to ``process_request`` -- a
+            mapping, or a sequence of name/value pairs.
+        spec: Which of them become config values. Worth naming: a client sends
+            headers of its own on every request (``host``, ``user-agent``, ...),
+            and ``unknown="reject"`` turns such a request away.
+        multi: Return every value of a repeated header as a list instead of
+            taking the first.
```
caterryan: **Say `multi` is unreachable on this engine.** The server hands the plugin a dict that keeps the last value of a repeated header or query parameter, so `multi=True` never sees a list here. It only matters for a runtime that delivers name/value pairs.

One sentence here, at line 474, and in the README. Otherwise an author will expect a list.

influxdata-plugin-utils/README.md
```diff
@@ -16,34 +16,154 @@ pip install -e influxdata-plugin-utils
...
+BODY = KeySpec(allowlist=["measurement", "field", "window"], unknown="reject")
+QUERY = KeySpec(allowlist=["window"])
+HEADERS = KeySpec(allowlist=["x-api-key"], rename={"x-api-key": "api_key"})
+ENV = KeySpec(allowlist=["PLUGIN_API_KEY"], rename={"PLUGIN_API_KEY": "api_key"})
+
+VALIDATORS = [
+    Validator("measurement", required=True),
+    Validator("api_key", required=True),
+    Validator("window", default="1h", cast=parse_timedelta),
+    Validator("limit", default=1000, cast=int, gte=1, lte=10_000),
```
caterryan, line 65: Use `parse_int` here, not `int`. JSON bodies now arrive typed, so `cast=int` turns `12.7` into `12` with no error. `parse_int` rejects it. Import it next to `parse_timedelta` on line 44. This is the example people will copy.

influxdata-plugin-utils/README.md
```diff
@@ -16,34 +16,154 @@ pip install -e influxdata-plugin-utils
...
+- `rename` maps a source key onto the config key it becomes;
+- `unknown` decides what happens to a refused key — `"ignore"` drops it,
+  `"reject"` names it in the error so the sender learns what was wrong.
+
+On a layer the caller controls, prefer `allowlist`: a parameter added to the
+plugin later stays unreachable until it is listed, where a `denylist` would let
+it through unnoticed.
+
+Header names are matched regardless of casing and hyphenation and become config
+keys (`X-Api-Key` → `x_api_key`). Everywhere else names are matched and kept
```
caterryan: Rewrite for the new header rule: names are matched regardless of case and become lowercased keys (`X-Api-Key` → `x-api-key`). Hyphens stay, so `rename` is how a plugin gets `api_key`.

Add one sentence: spell `rename` keys exactly as they appear in `allowlist`. The reachability check in `KeySpec.__post_init__` runs before names are lowercased, so `allowlist=["X-Api-Key"]` with `rename={"x-api-key": ...}` fails at construction.

influxdata-plugin-utils/README.md
```diff
@@ -16,34 +16,154 @@ pip install -e influxdata-plugin-utils
...
+exactly as written. `parse_env` requires an allowlist: the process environment
+belongs to the host and holds credentials, so nothing is read without being
+named. `parse_toml` refuses a path that does not name a `.toml` file before
```
caterryan: Add: `parse_toml` should only ever receive a path from a layer the operator controls (trigger arguments or the file itself), never from a request body, header or query string.

influxdata-plugin-utils/README.md
```diff
@@ -16,34 +16,154 @@ pip install -e influxdata-plugin-utils
...
+opening it; pass `require_suffix=False` for a config file named some other way,
+and `is_toml_path()` answers the same question without reading anything. `Authorization` never reaches a plugin — the engine authenticates with
+it — so a token needs a header of your own.
+
+A value that arrives empty — a blank string, a JSON `null`, an unset variable —
+is left out of its layer, so a validator default applies instead and a blank in
+one layer does not erase the layer below it. `0`, `False` and `[]` are real
```
caterryan, line 135: Add that header, query-string and environment values are trimmed of surrounding whitespace before anything else sees them (`_values_of` strips). The docstrings only say "strings".

influxdata-plugin-utils/CHANGELOG.md
```diff
@@ -21,6 +47,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  environment variables, the trigger arguments, the TOML file — and returns a
+  `Config`. A value that arrives empty is left out of its own layer, so a blank
+  trigger argument lets a validator default apply and a blank in the file no
+  longer erases the argument underneath it. Every failure is a `ValueError`,
+  including an unreadable file and a rejected value, and a `config_file_path`
+  that does not name a `.toml` file is now refused before the file is opened.
+  It stays supported, and `load_config` is the one to reach for in new plugins.
+- Configuration keys are stored as they arrive. Nothing in a layer is
+  interpreted, whatever a value spells, and a key is matched exactly as written
+  — except header names, which become config keys (`X-Api-Key` -> `x_api_key`)
```
caterryan: Header line: `X-Api-Key` -> `x-api-key` (lowercased only). Say that `-` is kept and that `rename` is how a plugin gets an underscore name.

influxdata-plugin-utils/CHANGELOG.md
```diff
@@ -21,6 +47,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  because their spelling comes from the protocol.
+
+### Removed
+
+- The `dynaconf` dependency. The package now has none.
```
caterryan, line 74: Add the departures from dynaconf here, so a ported plugin knows what changed:
- `must_exist` is removed; use `required`. (`must_exist=False` had meant "must be absent".)
- String defaults are no longer TOML-parsed, so `default="5", gte=1` now fails.
- A callable default is stored as-is, not called. (dynaconf called it with `(settings, validator)`.)
- Header names no longer fold `-` to `_`.

caterryan (commented):
Reviewed the config/validation rework against dynaconf 3.3.5 and the plugins in the repo. Requests inline, in short:

- bound the rejected value in validation error messages;
- remove `must_exist`, keep `required` (breaking, accepted);
- header names: lowercase only, stop folding `-` to `_`;
- README: use `parse_int` in the example, document header spelling, trimming, `parse_toml` trust, and `multi`;
- CHANGELOG: list the departures from dynaconf.

influxdata-plugin-utils/src/influxdata_plugin_utils/sources.py
```diff
@@ -0,0 +1,489 @@
...
+            duplicates = {name for name in targets if targets.count(name) > 1}
+            if duplicates:
+                raise ValueError(f"rename maps several keys onto {sorted(duplicates)}")
+            if self.allowlist is not None:
+                unreachable = set(self.rename) - set(self.allowlist)
+                if unreachable:
+                    raise ValueError(
+                        f"rename names {sorted(unreachable)}, which the allowlist "
+                        f"does not let through"
+                    )
```
caterryan, line 95: **Refuse a rename that lands on a denied key.** `KeySpec(denylist=["secret"], rename={"x": "secret"})` builds, and a body `{"x": 1}` then yields `{"secret": 1}`. Line 210 checks the denylist against the source key only, so the rename walks around it. Even `unknown="reject"` only catches a sender who also spells `secret` literally.

Add one check here, at construction: raise if any rename target is in the denylist.

The same spot can catch the allowlist mirror. `allowlist=["x-api-key", "x_api_key"], rename={"x-api-key": "x_api_key"}` is a guaranteed collision, since both source keys land on `x_api_key`, but today it only surfaces per request in `_claim`, and only when both keys arrive non-blank.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+    "not_startswith": (
+        lambda value, other: not value.startswith(other),
+        "must not start with {other}",
+    ),
+    "not_endswith": (
+        lambda value, other: not value.endswith(other),
+        "must not end with {other}",
+    ),
+    "regex": (
+        lambda value, other: re.search(other, value) is not None,
```
caterryan, line 82: **Anchor `regex` and `not_regex` at the start, as dynaconf does.** dynaconf uses `pattern.match`, so `regex="b"` rejects `"abc"` and `not_regex="b"` accepts it. `re.search` here flips both. Use `re.match(other, value)` on this line and on line 86 (`re.search` also accepts a compiled pattern, so nothing else changes).

The `not_regex=r"\s"` row in `tests/test_validation.py:88` only fails today because of the search semantics; see the comment there.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+    "eq": (lambda value, other: value == other, "must equal {other}"),
+    "ne": (lambda value, other: value != other, "must not equal {other}"),
+    "gt": (lambda value, other: value > other, "must be greater than {other}"),
+    "gte": (lambda value, other: value >= other, "must be at least {other}"),
+    "ge": (lambda value, other: value >= other, "must be at least {other}"),
+    "lt": (lambda value, other: value < other, "must be less than {other}"),
+    "lte": (lambda value, other: value <= other, "must be at most {other}"),
+    "le": (lambda value, other: value <= other, "must be at most {other}"),
+    "identity": (lambda value, other: value is other, "must be {other} itself"),
+    "is_type_of": (
```
caterryan, line 37: **Port dynaconf's `is_type_of`.** dynaconf (`validator_conditions.is_type_of`) handles `Union[...]` / `int | str`, `list[int]` and `tuple[int, ...]` element-wise, `dict[str, int]` key-and-value, a tuple of types, and resolves builtin names given as strings (`is_type_of="int"`, done in `_validate_names` via `__builtins__`). Plain `isinstance` raises on a parameterized generic and on a string, so `Validator("k", is_type_of=list[int])` is "cannot be checked" here for `[1, 2]` and `[1, "a"]` alike, where dynaconf passes the first and rejects the second. Same for `is_type_of="int"` on `5`.

Add rows for `list[int]` (pass and fail) and `"int"` next to `tests/test_validation.py:81`.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+                ("eq", eq),
+                ("ne", ne),
+                ("gt", gt),
+                ("gte", gte),
+                ("ge", ge),
+                ("lt", lt),
+                ("lte", lte),
+                ("le", le),
+                ("identity", identity),
+                ("is_type_of", is_type_of),
```
caterryan, line 158: **Run `is_type_of` first.** dynaconf reorders the operations so the type check runs before the others (`sorted(..., key=lambda item: item[0] != "is_type_of")`). In this fixed order the eight comparisons run first, so `Validator("k", gt=1, is_type_of=int)` on `"abc"` reports "cannot be checked with gt" instead of "must be of type int". Move this entry to the front of the tuple.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+        if self.cast is not None:
+            try:
+                value = self.cast(value)
+            except ValueError as exc:
+                raise ValueError(f"{name}: {exc}") from exc
+            except Exception as exc:
+                # a cast is the plugin's own code; its failure rejects the value
+                raise ValueError(f"{name}: cannot read {value!r}: {exc}") from exc
+
+        if self.condition is not None:
```
caterryan, line 226: **Run `condition` after the checks, not before.** dynaconf evaluates the operations first and the callable condition last. Here `Validator("k", gt=1, condition=...)` on `0` reports the condition's complaint where dynaconf reported "must gt 1". Move this block below the `for check, other in self.checks` loop (line 236).

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+    ``None`` counts as unset. A list, dict or set ``default`` is copied for
+    each use, so one rule's default cannot be changed through the values it
+    fills in. Checks are named explicitly, so a misspelled one is a
+    ``TypeError`` where the rule is written.
+    """
+
+    def __init__(
+        self,
+        *names: str,
+        required: bool | None = None,
```
caterryan, line 105: **Make `required` a plain bool.** In dynaconf `required` is the `must_exist` alias, so `required=False` means "must be absent" and `Validator("k", required=False)` rejects `{"k": "x"}`. Here it is no rule at all, the same trap the review already removed with `must_exist`. Declare it `required: bool = False`, and add the `required=False` spelling to the departures list in the CHANGELOG (comment on line 74 there).

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+    ):
+        if not names:
+            raise ValueError("a validator needs at least one key name")
+        self.names = names
+        self.required = bool(required)
+        self.default = default
+        self.apply_default_on_none = apply_default_on_none
+        self.cast = cast
+        self.condition = condition
+        self.when = when
```
caterryan, line 145: **Refuse a `when` that is not a `Validator` at construction.** dynaconf raises `TypeError("when must be Validator instance")` in `__init__`. Here `Validator("b", default=2, when=5)` constructs, and `validate()` then raises `AttributeError: 'int' object has no attribute 'holds'`, which is neither a construction-time error nor the `ValueError` the module docstring promises. Add:

```python
if when is not None and not isinstance(when, Validator):
    raise TypeError("when must be a Validator")
```

caterryan (commented):
Follow-up to my earlier review. I compared this `Validator` against dynaconf 3.3.5 check by check, running both through the same inputs (dynaconf via a settings object configured as the old `load_plugin_config` did, since that is what plugins actually ran). Everything below is a confirmed behavioural difference; message wording and the `ValidationError` -> `ValueError` class change are not listed on their own.

Code changes are requested where mirroring dynaconf is cheap: `regex` anchoring, `is_type_of` generics, check order, and two construction-time type checks. The deliberate departures stay as they are and go into the CHANGELOG list requested earlier. Existing plugins use only `default`, `cast`, `required`, `gt`, `gte`, `lte` and `is_in`, which behave identically on both sides.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+        not_regex=_MISSING,
+    ):
+        if not names:
+            raise ValueError("a validator needs at least one key name")
+        self.names = names
+        self.required = bool(required)
+        self.default = default
+        self.apply_default_on_none = apply_default_on_none
+        self.cast = cast
+        self.condition = condition
```
caterryan, line 144: **Refuse a non-callable `condition` at construction.** dynaconf raises `TypeError("condition must be callable")` in `__init__`. Here `Validator("k", condition=5)` constructs; it becomes a `ValueError` at validate time when the key is present and passes silently when it is absent. Add:

```python
if condition is not None and not callable(condition):
    raise TypeError("condition must be callable")
```

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
```diff
@@ -0,0 +1,256 @@
...
+        """Would this rule pass against ``values``? Used to answer ``when``.
+
+        A key nobody set does not hold: there is nothing to judge.
+        """
+        if self.default is _MISSING and any(
+            values.get(name, _MISSING) is _MISSING for name in self.names
+        ):
+            return False
+        try:
+            self.apply(dict(values))
```
caterryan, line 191: **No code change here, but three `when` behaviours depart from dynaconf and belong in the CHANGELOG departures list** (comment on line 74 there). All three are confirmed against dynaconf 3.3.5 through a settings object, as `load_plugin_config` ran it:

1. **An absent key does not hold** (line 189). dynaconf treats an absent key with no existence rule as passing, so `Validator("ripple", required=True, when=Validator("prototype", eq="cheby1"))` on `{}` raised "ripple is required" there and returns `{}` here; `{"ripple": 100}` with `lte=80` was rejected there and passes here. Documented in README:163 and the test at :159, so keep it, but say so.
2. **The `when` rule's own defaults and casts are not stored** (this line evaluates on a copy). dynaconf mutated settings inside `when`, so `Validator("k", lte=10, when=Validator("k", cast=int))` on `{"k": "5"}` passed there (`k` became `5`) and fails here (`lte` on a str). Test at :155 documents the copy.
3. **A `when` rule whose cast or condition raises counts as not holding** (line 192). dynaconf only caught `ValidationError`, so `when=Validator("a", cast=int)` with `a="x"` raised the `int()` error there and silently skips the dependent rule here. Test at :162 covers `cast=len` only.

influxdata-plugin-utils/tests/test_validation.py
```diff
@@ -0,0 +1,188 @@
...
+            (Validator("k", is_in=("mean", "max")), "median", "must be one of"),
+            (Validator("k", is_not_in=("admin",)), "admin", "must not be one of"),
+            (Validator("k", is_type_of=int), "5", "must be of type"),
+            (Validator("k", len_min=2), "a", "must be at least 2 long"),
+            (Validator("k", len_max=2), "abc", "must be at most 2 long"),
+            (Validator("k", contains="cpu"), "mem_usage", "must contain 'cpu'"),
+            (Validator("k", startswith="cpu"), "mem", "must start with 'cpu'"),
+            (Validator("k", endswith="_raw"), "cpu", "must end with '_raw'"),
+            (Validator("k", regex=r"^[a-z_]+$"), "CPU!", "must match"),
+            (Validator("k", not_regex=r"\s"), "two words", "must not match"),
```
caterryan, line 88: **Update once `regex`/`not_regex` use `re.match`.** `\s` does not match at the start of `"two words"`, so this row would stop failing. Use `" two words"` (leading space) or the pattern `r".*\s"`, and add a row that proves the anchoring: `(Validator("k", regex="b"), "abc", "must match")`.

influxdata-plugin-utils/tests/test_validation.py
```diff
@@ -0,0 +1,188 @@
...
+class TestChecks:
+    @pytest.mark.parametrize(
+        "rule, value, complaint",
+        [
+            (Validator("k", gt=0), 0, "must be greater than 0"),
+            (Validator("k", lte=10), 11, "must be at most 10"),
+            (Validator("k", ne="x"), "x", "must not equal 'x'"),
+            (Validator("k", is_in=("mean", "max")), "median", "must be one of"),
+            (Validator("k", is_not_in=("admin",)), "admin", "must not be one of"),
+            (Validator("k", is_type_of=int), "5", "must be of type"),
```
caterryan, line 81: Add rows for the generic and string forms once `is_type_of` is ported: `(Validator("k", is_type_of=list[int]), [1, "a"], "must be of type")` and `(Validator("k", is_type_of="int"), "5", "must be of type")`, plus a passing case for `list[int]` on `[1, 2]` in `test_passing_values_are_returned_unchanged`.

influxdata-plugin-utils/README.md
```diff
@@ -16,34 +16,162 @@ pip install -e influxdata-plugin-utils
...
+Validator("ripple", required=True, when=Validator("prototype", eq="cheby1"))
+```
+
+`required` asks for a usable value, so a key that arrives blank or `null`
+counts as unset. `when` applies a rule only while another one holds — and holds
+means the key is there and passes — while `condition` takes any predicate. The
+checks are `eq`, `ne`, `gt`, `gte`, `ge`, `lt`, `lte`, `le`,
+`identity`, `is_type_of`, `is_in`, `is_not_in`, `contains`, `cont`,
+`not_contains`, `len_eq`, `len_ne`, `len_min`, `len_max`, `startswith`,
+`endswith`, `not_startswith`, `not_endswith`, `regex` and `not_regex`. They are
```
caterryan, line 168: Add one sentence after the list: `regex` and `not_regex` match from the start of the value, as in dynaconf, so `regex="b"` rejects `"abc"` and a pattern that may appear anywhere needs `.*` in front.

influxdata-plugin-utils/CHANGELOG.md
```diff
@@ -21,6 +53,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `config.load_plugin_config` reads the same three layers as before — the named
+  environment variables, the trigger arguments, the TOML file — and returns a
+  `Config`. A value that arrives empty is left out of its own layer, so a blank
+  trigger argument lets a validator default apply and a blank in the file no
+  longer erases the argument underneath it. Every failure is a `ValueError`,
+  including an unreadable file and a rejected value, and a `config_file_path`
+  that does not name a `.toml` file is now refused before the file is opened.
+  It stays supported, and `load_config` is the one to reach for in new plugins.
+- Configuration keys are stored as they arrive. Nothing in a layer is
+  interpreted, whatever a value spells, and a key is matched exactly as written
```
caterryan, line 67: "A key is matched exactly as written" should also cover validator names: dynaconf matched `Validator("Rows")` to `rows` (case-insensitive) and walked a dotted name (`"a.b"`) into a nested dict. Here a rule name is matched exactly and `"a.b"` is a flat key. One clause here is enough; it is the other half of the same rule.

influxdata-plugin-utils/CHANGELOG.md
```diff
@@ -21,6 +53,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  — except header names, which become lower-case config keys (`X-Api-Key` ->
+  `x-api-key`)
+  because their spelling comes from the protocol.
+
+### Removed
+
+- The `dynaconf` dependency. The package now has none.
```
caterryan, line 74: **Extend the departures list I asked for earlier with what the check-by-check comparison against dynaconf 3.3.5 confirmed.** Together with the items already requested (`must_exist`, string defaults not TOML-parsed, callable defaults stored as-is, header names), a ported rule then has the whole picture:

- Every failure is a `ValueError`. dynaconf raised `ValidationError` (not a `ValueError`) and let anything but a `TypeError` escape raw from a cast, condition or check (`AttributeError` from `startswith` on an int, `re.PatternError` from a bad pattern, `KeyError` from a cast).
- `required=True` is not satisfied by `""`, whitespace or `None`, and it is checked before `cast`; dynaconf cast them (`cast=str` on `None` gave `"None"`).
- `required=False` is no rule; in dynaconf it meant "must be absent".
- `when`: an absent key does not hold, so the dependent rule is skipped where dynaconf ran it; the `when` rule's own defaults and casts are not stored; a `when` rule whose cast or condition raises counts as not holding instead of raising.
- A `set` default is copied per use; dynaconf shared the object, so mutating a result mutated the rule.
- Values keep their Python types: tuples stay tuples, dicts and lists are plain. dynaconf turned tuples into lists and handed `cast` its Box types (case-insensitive nested keys, attribute access), so `contains="A"` passed on `{"a": 1}` there.
- A misspelled check is a `TypeError` at construction; dynaconf raised `AttributeError` at validate time and passed silently when the key was absent.
- Not carried over: `env`, `messages`, `description`, `items_validators`, the `|` and `&` combinators, `validate_all`, `only`/`exclude`. A rule equal to an earlier one is applied, not dropped (dynaconf's `register` de-duplicated by equality, ignoring `default` and `apply_default_on_none`).

If the `regex` and `is_type_of` changes requested in `validation.py` land, they need no entry; otherwise add those two as well. No plugin in the repo uses `regex`, `when`, `is_type_of`, `condition`, `contains`, the `len_*` or `startswith` family, `required=False`, uppercase or dotted names, or a TOML-parsable string default, so this is about ported and future rules, not the existing ones.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
caterryan: Done in 7aa4fab: the parameter and the `bool(required or must_exist)` are gone, the test is removed, and the plugins moved to `required=True` in #165. Resolving.

influxdata-plugin-utils/src/influxdata_plugin_utils/sources.py
caterryan: Done in 3ba446f: `header_key` lowercases only, the `parse_request_headers` docstring and the tests now spell `x-api-key`, and a hyphen and an underscore are two headers. Resolving.

influxdata-plugin-utils/CHANGELOG.md
caterryan: Done: the Changed entry reads `X-Api-Key -> x-api-key`, and the new Added bullet says the name is otherwise kept as written and that `rename` gives another name. Resolving.

influxdata-plugin-utils/src/influxdata_plugin_utils/sources.py
caterryan: Partly. Refusing a repeat unless `multi=True` is a sound rule, and the docstrings and README describe it. Still missing is the sentence that the engine passes a plain dict, which can only carry one value per name, so on InfluxDB 3 itself a repeat never reaches the plugin and neither the refusal nor `multi` ever fires. Both only apply to a runtime that delivers name/value pairs. One sentence here and in the README paragraph. Leaving open for that.

influxdata-plugin-utils/README.md
caterryan: The header rule is done. Still missing: one sentence saying to spell `rename` keys as they appear in `allowlist`. `KeySpec.__post_init__` compares them before lowercasing, so `allowlist=["X-Api-Key"], rename={"x-api-key": ...}` still fails at construction on 7aa4fab. Leaving open for that.

influxdata-plugin-utils/README.md
caterryan: The paragraph moved, but the sentence is still missing: `parse_toml` should only ever get a path from a layer the operator controls, never from a request body, header or query string. Leaving open.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Fixed

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: fixed

influxdata-plugin-utils/src/influxdata_plugin_utils/sources.py
Aliaksei-Kharlap: fixed

influxdata-plugin-utils/README.md
Aliaksei-Kharlap: fixed

influxdata-plugin-utils/README.md
Aliaksei-Kharlap: added as a recommendation with the reason rather than a prohibition 
`"Take that path from a layer the operator controls — the trigger arguments, or the file itself. A path that arrives in the request body, a header or the query string lets the caller name any file the engine can read, and whatever parses as TOML becomes this plugin's configuration, another plugin's credentials included"  `
because three plugins — import, nori_regression and geo_enrichment — declare config_file_path among their request-body parameters. Example of use: several toml files have been created, and a user, via the API, passes only the path to the required file in the request body, depending on the situation. There might be a safety aspect to this approach worth discussing

influxdata-plugin-utils/README.md
Aliaksei-Kharlap: done

influxdata-plugin-utils/CHANGELOG.md
Aliaksei-Kharlap: Done

influxdata-plugin-utils/src/influxdata_plugin_utils/sources.py
Aliaksei-Kharlap: Fixed

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Fixed

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Fixed (It covers what dynaconf covered, minus three faults there: dynaconf accepts a tuple for list[int] and a list for tuple[int], reads tuple[int, str] as "any element of either type" rather than positionally, and raises on set[int] outright)

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Fixed

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Fixed

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Done

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Done

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: Fixed

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
Aliaksei-Kharlap: 1 - Done
2 - Done
3 - Fixed rather than documented: a `when` rule that cannot be evaluated (for example, due to type casting or predicate evaluation) is no longer considered "does not hold"; instead, an error is recorded for the rule that triggered the check, appearing as follows: `window: condition could not be checked: rows: invalid literal for int() with base 10: 'abc'`.

influxdata-plugin-utils/tests/test_validation.py
Aliaksei-Kharlap: Done

influxdata-plugin-utils/tests/test_validation.py
Aliaksei-Kharlap: Done

influxdata-plugin-utils/README.md
Aliaksei-Kharlap: Done

influxdata-plugin-utils/CHANGELOG.md
Aliaksei-Kharlap: Done

influxdata-plugin-utils/README.md
Aliaksei-Kharlap: Done

influxdata-plugin-utils/README.md
caterryan: The wording works, and it is better than a prohibition given that three plugins do take `config_file_path` from the body. Resolving this thread.

On the safety point: yes, it is worth its own issue. Anyone holding a database token can point those three plugins at any `.toml` the engine can read, and that file then becomes the plugin's configuration. The fix is per plugin, not here: pin `config_file_path` to the trigger arguments with `merge_config_layers(..., pinned=["config_file_path"])`, or drop it from the body allowlist. That is a separate PR.

influxdata-plugin-utils/src/influxdata_plugin_utils/validation.py
caterryan: Verified on 72fe3d6. Points 1 and 2 are in the CHANGELOG. Point 3 as a fix is the right call: `_Refused` separates "judged and said no" from "could not judge", `holds` only swallows the former, and the guarded rule reports the guard's failure with both names. A guard that plainly does not hold still lets the rule wait, per the new test. Resolving.

caterryan (approved):

PR #168, opened 2026-09-15
release: influxdata-plugin-utils 0.4.1

caterryan:
Ships the 0.4.0 changes as 0.4.1. `utils-v0.4.0` was tagged on `a4188a6` but never published: the publish action's bundled twine rejected core metadata 2.5, fixed in #166. The tag ruleset blocks moving `utils-v0.4.0`, so the release goes out under a new version rather than an admin bypass.

**Changes.**

- `__version__` 0.4.0 → 0.4.1. No source changes; the package contents are identical to what `utils-v0.4.0` would have built.
- `CHANGELOG.md`: the `## [0.4.0]` section is retitled `## [0.4.1] - 2026-09-15` with a short note at the top explaining why 0.4.0 has no release. The `[Unreleased]` and `[0.4.1]` compare links are updated, and the `[0.4.0]` link is dropped since nothing references it. This keeps the GitHub release notes, which the workflow builds from this section, carrying the actual feature list instead of pointing at a version that does not exist on PyPI.

**Verified locally.**

- Replayed the workflow's "Verify tag matches package version" step with `GITHUB_REF_NAME=utils-v0.4.1`: passes.
- Replayed "Generate release notes" for 0.4.1: section found, 115 lines, starts with the note and the `### Added` list.
- Test suite: 236 pass on Python 3.14. One failure, `test_a_deeply_nested_body_stays_within_the_valueerror_contract`, fails identically on `main` with the same interpreter and is a Python 3.14 behaviour change, not related to this PR. The CI matrix is 3.11 and 3.13, where it passes.

**After merge.** Tag the merge commit on `main` and push the tag; the workflow runs at the tagged commit, which now has the v1.14.2 action pin:

```bash
git fetch origin main
git tag -a utils-v0.4.1 -m "influxdata-plugin-utils 0.4.1" origin/main
git push origin utils-v0.4.1
```

`influxdata/geo_enrichment` declares `influxdata-plugin-utils>=0.4.0`, which nothing on PyPI satisfies until this ships.


caterryan:
Superseded: 0.4.0 will be released by moving `utils-v0.4.0` to `65daf64` (the #166 merge commit) rather than shipping a new version number, so this bump is not needed. Branch kept until the 0.4.0 release completes.

PR #170, opened 2026-09-15
refactor(geo_enrichment): load configuration through the utils 0.4.0 pipeline

caterryan:
## Summary

Rewrites the geo_enrichment configuration layer on top of the `config`, `sources` and `validation` modules of `influxdata-plugin-utils` 0.4.0. It is the first plugin to use them, so the shape follows the utils README: one module-level `VALIDATORS` list, one parser per source under a `KeySpec`, layers merged lowest precedence first with `load_config`.

- **Validators carry the rules.** Defaults, casts (`parse_int`, `parse_bool`, `parse_delimited_list`, a `keyword` cast for choices), and checks (`is_in`, `gt`, `gte`, `lte`, `len_min`) live in `VALIDATORS`. `when` rules express dependencies: `reference_file` for polygon/nearest, `target_measurement` for tag output, `priority_attribute` for the priority policy, `end` with `start`.
- **Sources are parsed in precedence order.** `parse_trigger_args` → `parse_toml` (the file wins) for the write trigger. `derive_config()` keeps only the cross-setting rules whose message has to name a second setting (target ≠ source, one coordinate input, grid precision range per grid, `nearest_count` per strategy) and computes the derived values.
- **Body keys are allowlisted with `unknown="reject"`**, so a typo comes back as a 400 naming the key.
- **Sentinels removed.** `max_radius_m` and `grid_precision` no longer default to `-1`; unset means the strategy's / grid's own default.
- Blank values mean "not provided", so a validator default applies.

## ⚠️ Behaviour change on the HTTP endpoint

The utils README is explicit that a TOML path must come from a layer the operator controls, never the request body. Following that:

- **The HTTP trigger now reads its trigger arguments** and the TOML file they name, exactly like the write trigger. Before, they were ignored with a warning.
- **What the operator sets there is pinned.** The body fills in whatever is left open plus the five backfill fields, but a body that names a pinned setting is refused with a 400 (`Cannot override pinned keys [...]`) rather than run with a configuration its sender did not ask for. Values for the backfill fields on the trigger act as defaults the body overrides.
- **`config_file_path` is no longer accepted in the body** (400: `Request body may not set 'config_file_path'`). Name the file in the trigger arguments instead.

Callers that sent `config_file_path` in the body need to move it to the trigger. A trigger created without arguments behaves as before: the body carries everything.

Error messages for validation failures now come from utils (`strategy must be one of (...)`, `target_measurement is required`, `batch_size: Invalid integer: 'many'`). README, TOML template, docstring metadata (`http_args_config` added, `config_file_path` dropped from `http_body_config`) and troubleshooting entries are updated to match. Version bumped to 0.2.0; both earlier registry versions were yanked because they were written against the dynaconf-era 0.4.0 pre-release.

## Test plan

- [x] `pytest test_geo_enrichment.py` — 189 passed against the published `influxdata-plugin-utils==0.4.0` wheel in a clean venv (byte-identical to the source on main).
- [x] New tests: pinned settings refuse a body override and accept a body that leaves them alone; `config_file_path` in the body is a 400; backfill defaults on the trigger are overridden by the body; the config file overrides trigger arguments; blank arguments leave defaults in place; keyword settings ignore case and whitespace; unknown body keys are a 400; the docstring metadata is tied to `SETTING_KEYS` / `CONFIG_KEYS`.
- [x] `influxdb3-plugin validate` (SDK v0.6.0, the version CI pins) passes; `package` produces `geo_enrichment-0.2.0.tar.gz`.
- [x] `scripts/validate_readme.py --plugins geo_enrichment` reports the same two pre-existing template errors as on main; nothing new.
- [ ] Exercise a write trigger and a backfill against a live InfluxDB 3 with a config file named on the HTTP trigger.


influxdata/geo_enrichment/geo_enrichment.py
```diff
@@ -177,11 +177,11 @@
             "required": false
         }
     ],
-    "http_body_config": [
+    "http_args_config": [
         {
             "name": "source_measurements",
```
Aliaksei-Kharlap, line 182: http_args_config marks source_measurements and output_columns as required: true—and those exact same two fields are marked as required: true in http_body_config. In Explorer, the user will be required to specify them as arguments when creating the HTTP trigger, even though the README states: "Without trigger arguments the request body carries the whole configuration."

influxdata/geo_enrichment/geo_enrichment.py
```diff
@@ -1180,163 +1358,191 @@ def memo_store(memo: OrderedDict, key: tuple, value, cache_size: int) -> None:
...
+    # unset means the grid's own default; the range depends on the grid as well
+    Validator("grid_precision", cast=parse_int, gte=0, lte=30),
+    Validator("unknown_value", default="UNKNOWN", cast=str),
+    Validator("quantize_decimals", default=4, cast=parse_int, gte=0, lte=9),
+    Validator("cache_size", default=100_000, cast=parse_int, gte=1),
+    # what one setting demands of another
+    Validator("target_measurement", required=True, when=Validator("output_mode", eq="tag")),
+    Validator("reference_file", required=True, when=Validator("strategy", is_in=REFERENCE_STRATEGIES)),
+    Validator("priority_attribute", required=True, when=Validator("overlap_policy", eq="priority")),
+    # the backfill window and switches, read by the HTTP endpoint alone
```
Aliaksei-Kharlap, line 1451: Five backfill fields are now validated on the write trigger as well. The list of validators was shared previously, and cfg["backfill"] was constructed at both entry points—but these fields had only a single rule: Validator(name, default=""). There was no casting or checking, so they couldn't trigger a failure; the actual parsing (int(), parse_bool, start/end pairing) happened later, in process_request. Now, the parsing logic has moved into the validation rules, and validate() runs the entire list for both entry points—handling values ​​that previously made it silently into cfg["backfill"] and sat there unprocessed

influxdata/geo_enrichment/geo_enrichment.py
```diff
@@ -177,11 +177,11 @@
...
         }
     ],
-    "http_body_config": [
+    "http_args_config": [
         {
             "name": "source_measurements",
-            "example": "gps",
-            "description": "Table to backfill. One call handles one table: if several are given, the first is used and the rest are ignored.",
+            "example": "gps fleet_pos",
+            "description": "Space-separated source tables to enrich. Rows from other tables in the batch are ignored.",
```
Aliaksei-Kharlap, line 184: This is not the case for HTTP—only the first table is used, and the others are discarded with a warning

influxdata/geo_enrichment/geo_enrichment.py
```diff
@@ -1180,163 +1358,191 @@ def memo_store(memo: OrderedDict, key: tuple, value, cache_size: int) -> None:
...
+
+    The body has the last word, except on a setting the operator fixed on the
+    trigger: that one it may not name at all, so a request is refused rather
+    than run with a configuration its sender did not ask for. The backfill
+    window and switches are always the body's; a trigger value is the default.
+    """
+    layers = merge_config_layers(
+        operator_settings(args or {}),
+        parse_json_body(request_body, REQUEST_BODY),
+        pinned=SETTING_KEYS,
```
Aliaksei-Kharlap, line 1500: As I understand it, for an HTTP trigger, it is currently impossible to override any parameter via the request body if it is already defined in `args` (with the exception of `BACKFILL_KEYS`). I have a quick question: should we consider this the preferred behavior for our plugins?

Aliaksei-Kharlap, line 1500: A point for discussion: perhaps we should add pinned and on_conflict parameters to load_config as well?

caterryan:
Superseded by https://github.com/influxdata/influxdb3_plugins/pull/171. The branch was rewritten so the request body overrides operator-set settings instead of being refused by them, with the review fixes as individual commits.

